PR #2, opened 2015-12-03
Tidy up apt-get and remove redundant packages, using @oletterers work.

swizzlr:
This PR pares down the packages needed to the bare essentials.


hamin:
merging

Danappelxx:
What's the point of `rm -rf /var/lib/apt/lists/*`? After that line I can't even run `apt-get install curl`.

PR #18, opened 2015-12-18
Change base image to ubuntu.

swizzlr:
Tentatively suggesting we do this, and given what we've learned about dockerizing swift, it would be trivial to create a phusion flavour (but I would also suggest that process forking should go the way of the dodo...).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:14.04
 MAINTAINER Haris Amin <aminharis7@gmail.com>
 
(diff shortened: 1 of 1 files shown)


hamin:
@swizzlr can we specify the 14.04 image from ubuntu? I'm ok not using phusions

swizzlr:
Agreed.

hamin:
@swizzlr so does this PR require 14.04? Do we need to specify that in the 'FROM' in docker file?

PR #20, opened 2016-01-31
Update to the January 25th development snapshot

briancroom:
I've been happily using this as a dev environment for the past week.

Note that this introduces the concept of multiple branches/tracks of snapshots which is new as of the latest set of archives. (See https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20160125/000934.html)


hamin:
@briancroom thank you so much. This looks great. Especially the branches/tracks bit. I'd missed that from the mailing list.

PR #21, opened 2016-02-25
Update to Feb. 25th snapshot

briancroom:
There's another new snapshot out! This one includes `swift test` for the first time :grinning:


hamin:
@briancroom thank you :) and this is exciting :)

Issue #9, opened 2015-12-03
Update documentation to clarify REPL usage and security implications


swizzlr:
## Overview

Swift REPL requires LLDB. LLDB requires some elevated privileges.

## Objectives

- [ ] Clearly define the elevated privileges required by LLDB's REPL
- [ ] Provide clear and simple documentation in the README on how to access the REPL, including an explanation of the rationale required
- [ ] Provide alternative examples of using the docker image that don't involve the REPL/elevated privileges (e.g. Dockerfiles, invoking scripts in Swift)

hamin:
The real was definitely working for me, but I was installing all the deps that apple was suggesting. Should we put those deps back for completeness sake? We might be breaking other parts of the toolchain without knowing it but not having deps apple details in their docs. Perhaps revert that commit and see what happens?

swizzlr:
@hamin When they say "development dependencies" I assume they mean for building the swift toolchain itself. (right, @jckarter)? I'll work backwards from the list of dependencies and find which libraries are necessary; then I'll open a PR to the swift repo updating their docs.

hamin:
@swizzlr yeah what you're saying makes sense, checking it out

swizzlr:
@hamin just tried "reverting" those changes and no dice. are you sure it worked for you, back then? maybe it's flaky?

hamin:
So its working for me right now. I'm running latest master:

![screen shot 2015-12-03 at 7 06 51 pm](https://cloud.githubusercontent.com/assets/28212/11578183/13091f66-99f1-11e5-98f4-fdc24e76e96f.png)

hamin:
@swizzlr its working for me. I'm running latest master

hamin:
@swizzlr what were you trying to run exactly?

hamin:
@swizzlr argh sorry its not running for me...i was running my old docker image...gimme a few

hamin:
@swizzlr confirmed this master is not working for me...the above screenshot was from my original repo. It definitely worked

swizzlr:
Does this commit work 369c495ac18df95d9cf48df22a71df299446cd2c ?

hamin:
@swizzlr checking it out. Meanwhile do a docker pull from here https://hub.docker.com/r/harisamin/docker-swift/

`docker pull harisamin/docker-swift`

Try that and see if it works there

swizzlr:
Nope, not working on this (digital ocean VM hooked up with docker-machine).

```
[...]

PR #24, opened 2016-04-10
[BUG] REPL doesn't work

gabhi:
Updated documentation for #9


hamin:
@gabhi The REPL seems to be working again w/ the latest May snapshot. See #26 . Is the privileged option still recommended?

hamin:
@gabhi seems like this is still an issue and your doc update fixes it, thanks :)

PR #25, opened 2016-04-20
Fixes for April 12 Snapshot

rtsisyk:


hamin:
Hey @rtsisyk thanks for the PR. We would actually like to keep the base image to 14.10 since its the last Ubuntu release with Long Term Support (LTS). Of course other images can override those ENVs based on this docker file. I did see that you have a fix for broken file permissions. Would you care to submit a PR with just the file permissions fix? Thanks again for your help. Its highly appreciated :)

hamin:
@rtsisyk This issue seems to be have resolved in the latest May snapshot #26 . Thank you though :)

PR #27, opened 2016-06-07
use 2016-06-06 snapshot

bppr:
As you can see from the [snapshots](https://swift.org/download/#snapshots) section on Swift.org, there's a new release!


hamin:
@bppr thank you :)

PR #33, opened 2016-09-21
Use clang-3.6 as compiler on Trusty

zwaldowski:
Clang 3.6 should be set as the compiler for Swift to use the preferred Gold linker.

See also:
- https://github.com/swiftdocker/docker-swift-dev/blob/master/Dockerfile#L26
- https://bugs.swift.org/browse/SR-2299


hamin:
@zwaldowski awesome thank you very much

PR #35, opened 2016-11-16
Allow SwiftPM to function inside image

zwaldowski:
Installs `curl` in the image so that `swift build` and `swift test` work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ENV SWIFT_PLATFORM ubuntu16.04
 
 # Install related packages and set LLVM 3.6 as the compiler
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


ChasonTang:
+1

hamin:
@zwaldowski thanks just saw this :). Hmm so I'm completely fine merging this...just wondering how I should deal with the existing tags for `3.0.0` and `3.0.1` on docker hub. Thoughts?

zwaldowski:
@hamin Not an expert, but I see no harm in re-tagging? Makes it complicated for 3.0.0, not sure what to do there.

hamin:
@zwaldowski ok merged and there's a new `3.0.1` tag on Docker hub. Unfortunately `3.0.0` still has that bug. But I think its ok for now :). Thanks for the PR @zwaldowski

PR #38, opened 2016-12-13
<Description> : Adding vim as an editor to allow prototyping with swift.

Khalian:
<Type> : feature

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ENV SWIFT_PLATFORM ubuntu16.04
 
 # Install related packages and set LLVM 3.6 as the compiler
 RUN apt-get update && \
-    apt-get install -y build-essential wget clang-3.6 curl libedit-dev python2.7 python2.7-dev libicu-dev rsync libxml2 git libcurl4-openssl-dev && \
+    apt-get install -y build-essential wget clang-3.6 vim curl libedit-dev python2.7 python2.7-dev libicu-dev rsync libxml2 git libcurl4-openssl-dev && \
     update-alternatives --install /usr/bin/clang clang /usr/bin/clang-3.6 100 && \
     update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 100 && \
     apt-get clean && \


swizzlr:
Vi comes preinstalled with Ubuntu, so I'm not sure this is needed?

Khalian:
I tried using vi after spinning up a container from this image. I could not find it in /usr/bin or in $PATH. After this change I was able to successfully use vim. 

I effectively did not want anyone else to go through this issue again hence the PR.

swizzlr:
@Khalian I feel that we should try to keep this repo as focussed and lightweight as possible, but I also think that having an editor preinstalled is a great convenience. Maybe this is something that should be added as a customisation in your own Dockerfile; or maybe this is something that you would like to maintain on Docker Hub as a swiftdocker `development` repo, with useful tools for debugging and developing?

swizzlr:
Or we could just add `vim` because I mean it's not a big deal.

Khalian:
When I first wanted to try out this image, I wanted to prototype swift in Linux. I wager that's the expectation from most devs. However there wasn't a single text editor in the system. I had to rebuild the image with vim to actually make this image usable to me.This was a completely broken user experience. As a developer I put accessibility and user experience as a premium over optimization. It's upto you if you think users will be OK without atleast one text editor.

hamin:
@Khalian @swizzlr I'd like to keep the image size small. In fact would like to do more work on this to make it smaller. I don't see the need to add vim as baseline. You can always make a Dockerfile that that uses this swift docker image as a base image and added Vim. That's very simple to do. Happy to give an example if you need it.

jatoben:
@Khalian Why not just bind-mount your source directory into the container? That's how I use this for CI on Linux — see https://github.com/jatoben/RocksDB/blob/master/.travis.yml as an example.

Khalian:
@swizzlr I have lost count on how many developers do not get the principle accessibility and try to be clever(like @jatoben is doing) trying to optimize the wrong thing. If you do not provide accessibility for the right use case, your product is pointless. Note I did not install a full blown IDE for the very purpose of not bloating up the image despite my urge to do so(my original intent was to include a pathogen swift autocomplete plugin but I decided against it for the very purpose of efficiency). But I am livid folks are fighting over a goddamn text editor (a modal hyper efficient one at that) while you have the entire userspace of ubuntu on it. This is precisely why everyone believes software sucks, this attitude of being clever permeates that being clever ends up making software inaccessible to those that matter - the end user. 

Never try to optimize over something that creates a broken user experience. My requirement was simple, have a sandbox on my macbook that allows me to to quickly test out scripts. I can very well create an offshoot of this project and add what I want, I submitted this PR to you so that folks who consume what appears to be the canonical swift docker image do not have to spend even a nanosecond of their time on things that should be a given(and yes everyone assumed vi is a given because the base image is ubuntu (please do not pretend otherwise), which did not actually reflect reality). 

This whole thread is very similar to the debates on whether i++ is better than ++i. @hamin right now the argument is between 1 million bytes and 1 million and 1 bytes(that was a figure of speech)

swizzlr:
I'm merging this on the basis that it does very little harm and honestly I can sympathise with the frustration that people experience when starting out.

All of this being said, @Khalian, I don't know if it was intended but your comments seemed quite aggressive. Haris has put a lot of time and effort into maintaining this repository, as have we all generously taken the time to seriously and objectively consider your proposal; it is very unproductive to let this descend into complaints about the quality of the discourse. Please be nicer to OSS developers or I'll make the base image [Alpine Linux](https://hub.docker.com/_/alpine/) 😈

hamin:
@Khalian let's talk about accessibility. Maybe you're new to Docker community. I'm by no means an expert and am still learning. Having smaller images in Docker context has a very DIRECT impact on accessibility and usability IMHO. I've yet to really evaluate how much smaller our image can get. Most of it depends on the great folks Apple and Swift core team who work tirelessly to improve the Swift community. 

Most ppl in the Docker community are trying to get their image sizes small for deploying applications in docker environments. That's a direct accessibility + developer experience issue. Surely you can understand and respect our wishes to try to achieve that? Looking at the image size of the builds on Docker hub https://hub.docker.com/r/swiftdocker/swift/tags (comparing the size of the latest build vs the 3.0.2 tagged build) I can see that adding Vim added 12MB to the image size. Maybe this is peanuts but again the we must ask ourselves is this really needed in a base docker image for Swift that provides a sufficient dev+deployment environment? 

> This whole thread is very similar to the debates on whether i++ is better than ++i

Yet here you are trying to enforce one editor over another one. Do you see the problem in your argument? Have you heard that vim/emacs/nano/any other text editor flame wars exist in our world? Why even open that door when you're more than free to add whatever text editor you want to your environment? And for the record, I myself use nano. Maybe you think I'm crazy, but I'm not forcing it on you. I have friends that use emacs on ubuntu. Good for them. To each their own. And that's absolutely fine. I understand you don't want to waste ppl's time for even a nanosecond on what editor they want to use. But who are you (or really any of us) to judge or enforce that on them. I'd rather save them perhaps some microseconds/seconds on 12 additional Megabytes when they're deploying their app in a Docker environment.

@jatoben was actually trying to give you some great advice and instead you berated him too. Especially when working on a mac, being able to just share your source directory to your container is way easier. You make the changes on your mac in whatever editor you desire (Xcode/Sublime Text/Visual Studio Code/Atom/Vim/Emacs) and just build your container again. I think the Docker docs could be a bit clearer, and perhaps our docs could be better. But @jatoben was nice enough to share a very nice and easy to try solution that doesn't take much effort at all, instead you insult him just as you insulted me.

Which brings me to this:

>This is precisely why everyone believes software sucks, this attitude of being clever permeates that being clever ends up making software inaccessible to those that matter - the end user.

We are thinking of the end user, and in case of Docker images, the end user is ALSO the person deploying their applications in a docker environment who cares or at least appreciates slimmer base images. But more importantly, one of the main reasons IMHO that people believe that "software sucks" and more specifically why "open source software sucks" is precisely because of ppl like you @Khalian and your lack of respect in rants/comments like this. Would you really speak that way to me or @jatoben or @swizzlr if were right there in person? Do you really think your tone and comments toward us are warranted for a "product" that we care to support in our own free time so that others can benefit? Do you honestly think this is the way to behave at all?

@swizzlr is awesome and accepted the PR. So you won. I too would rather not spend an ounce more of energy trying to make you happy. I can't promise if this stays, because as I said, I would like to spend time reducing the image size significantly if I can. But maybe I won't get to it,  as you must realize, I do this on my OWN FREE time. Maybe someone who doesn't insult me or anyone else's efforts to make this project better and help ppl who are using it can bring up better arguments that you have put so forth. @Khalian you're an engineer/developer yourself. I just had a long hard week of work, and was greeted by your latest comment on this thread. Surely you can understand how rude and unnecessarily a rant/insult such as your last comment is?

No-one here owes you anything, but perhaps...just perhaps @Khalian you owe the ppl in this conversation an apology. 

@jatoben @swizzlr sorry to mention you guys in this comment, but I felt it necessary to address this.

A few words of advice @Khalian ... don't be mean.

PR #43, opened 2016-12-30
Documentation: Official Docker Image

Khalian:
<Type> : docs

### DIFF
--- a/docs/LICENSE.md
+++ b/docs/LICENSE.md
@@ -1,0 +1,7 @@
+Copyright (C) 2017 by Swiftdocker maintainers/contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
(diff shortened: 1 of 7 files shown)


Khalian:
https://github.com/swiftdocker/docker-swift/issues/40

docs/LICENSE.md
```diff
@@ -0,0 +1,7 @@
+Copyright (C) 2015 by Haris Amin
```
swizzlr: 2017 – Copyright attribution should perhaps be "Swiftdocker maintainers"? @hamin ?

docs/README-short.txt
```diff
@@ -0,0 +1 @@
+An Ubuntu docker image containing the latest version of swift.
```
swizzlr: Since this description is really meant to be more about the product itself, perhaps we could steal the following from the official swift website:

> Swift is a general-purpose programming language built using a modern approach to safety, performance, and software design patterns.

https://swift.org/about/

swizzlr (changes requested):
This is a great start, thank you so much for contributing this Khalian. I'm going to take a knife to the README, if you don't mind, since we need it to be a bit more general purpose.

docs/README-short.txt
swizzlr: Needs to be less than 100 chars. 

> Swift is a general-purpose programming language using a modern approach to safety and performance.

docs/content.md
```diff
@@ -0,0 +1,43 @@
...
+##### Pull the Docker Image From Docker Hub:
+
+```bash
+docker pull swift
+```
+
+##### Create a Container from the Image and Attach It:
+
+```bash
+docker run --privileged -i -t --name swiftfun swift:latest /bin/bash
```
swizzlr: Docker won't accept us recommending --privileged. Let's add a note describing the problem. Hopefully someone else might fix it one day (short version: REPL depends on lldb which needs privileged kernel access or something)

swizzlr:
Wooooo https://github.com/docker-library/official-images/pull/2500

docs/LICENSE.md
Khalian: I would let you guys decide on what exact string should be here.

docs/README-short.txt
Khalian: Done

docs/content.md
Khalian: Ok so a comment for the same ?

docs/LICENSE.md
hamin: @swizzlr I've seen some other projects have a Contributors.MD with he list of the contributors. But Yeah I'm cool with "Swiftdocker maintainers/contributors"

Khalian: Done.

docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+##### Pull the Docker Image From Docker Hub:
+
+```bash
+docker pull swiftdocker/swift
+```
+
+##### Create a Container from the Image and Attach It:
+
+```bash
+docker run --privileged -i -t --name swiftfun swiftdocker/swift:latest /bin/bash
```
tianon: Any particular reason you're including `--privileged` here?  That's essentially giving the container full unfettered access to your host, so definitely not recommended. :sweat_smile:

Khalian: I plagiarized the README from the project.

swizzlr: It's a bug with Swift - only required if you need a REPL, which is an uncommon requirement anyway so we'll remove it. I'll improve this today and submit @tianon

swizzlr:
Official PR https://github.com/docker-library/docs/pull/794

PR #46, opened 2017-01-12
BIG CHANGES

swizzlr:
@hamin @tianon

I took this opportunity to substantially upgrade the Dockerfile.

- Improved formatting of package install (hard wrapped errything)
- Key importing, download and install happen in one layer (as it should be)
- Suppress output of apt and GPG, which seem to enjoy printing to STDERR when there's no error. The build is now effectively silent.

This checks all the boxes!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ RUN apt-get -q update && \
     make \
     libc6-dev \
     clang-3.6 \
-    vim \
     curl \
     libedit-dev \
     python2.7 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,10 +16,10 @@ RUN apt-get -q update && \
     rsync \
     libxml2 \
     git \
-    libcurl4-openssl-dev && \
-    update-alternatives --quiet --install /usr/bin/clang clang /usr/bin/clang-3.6 100 && \
-    update-alternatives --quiet --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 100 && \
-    rm -r /var/lib/apt/lists/*
+    libcurl4-openssl-dev \
+    && update-alternatives --quiet --install /usr/bin/clang clang /usr/bin/clang-3.6 100 \
+    && update-alternatives --quiet --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 100 \
+    && rm -r /var/lib/apt/lists/*
 
 # Everything up to here should cache nicely between Swift versions, assuming dev dependencies change little
 ENV SWIFT_BRANCH=swift-3.0.2-release \
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM ubuntu:16.04
 MAINTAINER Haris Amin <aminharis7@gmail.com>
 
 # Install related packages and set LLVM 3.6 as the compiler
-RUN apt-get -qq update && \
-    apt-get -qq install -y \
+RUN apt-get -q update && \
+    apt-get -q install -y \
     make \
     libc6-dev \
     clang-3.6 \
@@ -16,7 +16,7 @@ RUN apt-get -qq update && \
     rsync \
     libxml2 \
     git \
-    libcurl4-openssl-dev > /dev/null 2>&1 && \
+    libcurl4-openssl-dev && \
     update-alternatives --quiet --install /usr/bin/clang clang /usr/bin/clang-3.6 100 && \
     update-alternatives --quiet --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 100 && \
     rm -r /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,29 +2,56 @@ FROM ubuntu:16.04
...
+      # uid                  Swift 2.2 Release Signing Key <swift-infrastructure@swift.org>
+          1BE1E29A084CB305F397D62A9F597F4D21A56D5F \
+      # pub   4096R/91D306C6 2016-05-31 [expires: 2018-05-31]
+      #       Key fingerprint = A3BA FD35 56A5 9079 C068  94BD 63BC 1CFE 91D3 06C6
+      # uid                  Swift 3.x Release Signing Key <swift-infrastructure@swift.org>
+          A3BAFD3556A59079C06894BD63BC1CFE91D306C6 \
+        ; do \
+          gpg --quiet --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" > /dev/null 2>&1; \
+        done \
+    && gpg --batch --verify --quiet swift.tar.gz.sig swift.tar.gz  > /dev/null 2>&1 \
```
tianon: Same note about `2>&1` here.

Dockerfile
```diff
@@ -2,29 +2,56 @@ FROM ubuntu:16.04
...
+      # pub   4096R/21A56D5F 2015-11-28 [expires: 2017-11-27]
+      #       Key fingerprint = 1BE1 E29A 084C B305 F397  D62A 9F59 7F4D 21A5 6D5F
+      # uid                  Swift 2.2 Release Signing Key <swift-infrastructure@swift.org>
+          1BE1E29A084CB305F397D62A9F597F4D21A56D5F \
+      # pub   4096R/91D306C6 2016-05-31 [expires: 2018-05-31]
+      #       Key fingerprint = A3BA FD35 56A5 9079 C068  94BD 63BC 1CFE 91D3 06C6
+      # uid                  Swift 3.x Release Signing Key <swift-infrastructure@swift.org>
+          A3BAFD3556A59079C06894BD63BC1CFE91D306C6 \
+        ; do \
+          gpg --quiet --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" > /dev/null 2>&1; \
```
tianon: Same note about `2>&1` here.

Dockerfile
```diff
@@ -2,29 +2,56 @@ FROM ubuntu:16.04
...
+    vim \
+    curl \
+    libedit-dev \
+    python2.7 \
+    python2.7-dev \
+    libicu-dev \
+    rsync \
+    libxml2 \
+    git \
+    libcurl4-openssl-dev > /dev/null 2>&1 && \
```
tianon: If this fails, `2>&1` will swallow the error message, right? (makes debugging failures somewhat harder -- I'd rather have easy debugging and a verbose build than a quiet build that's difficult to debug)

tianon (commented):
A few notes about debuggability, but otherwise looks sane!

Is `vim` used by the Swift core, or just something that's useful to have in the container? (might be worth leaving that out -- didn't notice it before)

Dockerfile
swizzlr: You're right, I'm not wedded to it. Better to be able to debug later, most people will consume this as an image, not by building manually.

Dockerfile
swizzlr: Re: Vim – not at all required, but having some text editor available was requested by users, and we didn't think it had a significant impact on image size. Would it be better for us to remove it?

swizzlr:
Oh, and I got to remove wget in favor of curl. 🍾

hamin:
@swizzlr @tianon so i'm ok with removing `vim`. However, I don't think we can remove `curl`. The reason is that `Foundation` uses curl for `NSURLSession`/`NSURLRequest` underneath. So I think we would still want that and also it's used for Swift's package manager. See #35

Dockerfile
```diff
@@ -2,29 +2,56 @@ FROM ubuntu:16.04
...
-    apt-get install -y make libc6-dev wget clang-3.6 vim curl libedit-dev python2.7 python2.7-dev libicu-dev rsync libxml2 git libcurl4-openssl-dev && \
-    update-alternatives --install /usr/bin/clang clang /usr/bin/clang-3.6 100 && \
-    update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 100 && \
-    rm -rf /var/lib/apt/lists/*
+RUN apt-get -q update && \
+    apt-get -q install -y \
+    make \
+    libc6-dev \
+    clang-3.6 \
+    vim \
```
tianon: Just tested in a bare `ubuntu:16.04`, and installing just `vim` is ~57.87 MB, which is space users can't ever reclaim if it's in the base image, whereas the following is pretty trivial:

```dockerfile
FROM swift
RUN apt-get update -qq && apt-get install -yqq vim && rm -rf /var/lib/apt/lists/*
```

Combine that with automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) and it's reasonably easy to have an up-to-date image built `FROM` this one with `vim` pre-installed.

hamin: @swizzlr what do u think?

swizzlr: @hamin @tianon agreed. great idea with repo links. let's do this.

tianon (approved):
:metal:

hamin:
Omg amazing thank you :) @tianon . Nice work @swizzlr

PR #59, opened 2017-01-25
22 Jan snapshot.

swizzlr:


hamin:
@swizzlr so should master be keeping track of snapshots? or should it leave the stable version?

swizzlr:
Oh shit yeah

Issue #55, opened 2017-01-21
Automated generation of new Dockerfiles and images


swizzlr:
Currently every new release requires manual generation of new Dockerfiles and images.

Continuous integration using Travis CI, the RSS feed from the releases page and the nightly snapshots would allow this to happen.

Anybody want to take this on?

Khalian:
Any tutorials on how to do this ?

hamin:
no tutorials... i think a while ago we figured that we could get some sort of notification from swift build releases...i think ppl on swift mailing list liked the idea, but i don't think we got a follow up on it

andradei:
We should get this going :)

florianreinhart:
Docker’s guidelines for official images advise against creating images from CI builds:

> Unlike a release candidate, continuous integration builds which have a fully automated release cycle based on code commits or a regular schedule are not appropriate.

https://github.com/docker-library/official-images#library-definition-files

Nevertheless, it’d be great to test current development snapshots of a specific version branch, e.g. Swift 4.2. Maybe these builds could be considered “release candidates”.

Can someone check this with the official images maintainers?

norio-nomura:
FYI, my repository([norionomura/swift](https://hub.docker.com/r/norionomura/swift/)) has been automated for Dockerfile update and image creation by using [CircleCI 's scheduled workflow](https://circleci.com/gh/norio-nomura/workflows/docker-swift).

andradei:
@florianreinhart If I understand that document correctly, _regular schedules_ aren't appropriate. How about some sort of automated script that checks whether a new version was released (not new commits, but full-fledged releases) and builds a new docker image?

florianreinhart:
It's OK to automatically generate a new image whenever a new Swift version is released.

I am just not sure about the Swift development releases generated from a development branch like [swift-4.2-branch](https://github.com/apple/swift/tree/swift-4.2-branch). These are generated by Swift CI almost every day.

shahmishal:
@swizzlr We are moving Swift Docker issues to [https://bugs.swift.org](https://bugs.swift.org) Component: Docker. Can you please file your issue on bugs.swift.org?

We are planning on closing GitHub swift-docker issues on Dec 20th, 2018 (PST).

Issue #50, opened 2017-01-13
Automated images based on new snapshots


swizzlr:
If we can get an RSS feed or something of new snapshots, we can auto push new snapshots for access to preview versions of Swift.

hamin:
@swizzlr that would be awesome...would make it way easier to play with latest snapshots...I think swift team is doing nightly snapshots now... but yes some api/rss feed or whatever would be great. Another thing to discuss with swift core community

swizzlr:
Whoever's running Swift CI could let us know – perhaps there's a jenkins plugin.

mackoj:
Hi, we can get all new snapshots from this feed https://github.com/apple/swift/releases.atom.

swizzlr:
Are those precompiled or just a source tree?

mackoj:
This rss feed represent each tag so it's only source code. If Apple provide precompiled version directly on Github you could access it with https://developer.github.com/v3/repos/releases/#get-a-single-release.

swizzlr:
Apple provide nightly snapshots from their Jenkins on the website. We need some way of retrieving the latest snapshot's URL/name.

aidansteele:
Each one of the snapshots listed on GitHub (on the Atom feed linked above) has a corresponding binary at `https://swift.org/builds/development/ubuntu1604/swift-$TAG/swift-$TAG-ubuntu16.04.tar.gz`. 

I'd be happy to submit some PRs to a) have a build arg to (default) build releases or build nightlies and b) automating the building and publishing of nightlies. Is that something that would be accepted?

hamin:
@aidansteele that sounds great but I think we're waiting to hear back from some folks at apple/swift core team to see if they can provide any deeper integration. @swizzlr correct me if I'm wrong

swizzlr:
@hamin true but tbh I doubt it's going to be much fancier than that. We can get going on this I think in the meantime.

@aidansteele submit away. How do you propose we automate this? Jenkins?

aidansteele:
@swizzlr I'm a big fan of Jenkins and use it for all my personal projects, but the problem is hosting. I'm not aware of any free hosted Jenkins services for OSS projects, e.g. like Travis. 

I would be more than happy to host this project's CI on the Jenkins server I'm using for http://github.com/awswift, but that would introduce a dependency on me and it'd make sense for others to be uncomfortable with that. So maybe Travis is the way ahead?

swizzlr:
[...]

PR #61, opened 2017-01-31
ARG instead of ENV for selecting Swift version

aidansteele:
I'm submitting this separately to the work needed for #50 / #55, but it will be helpful for those. Once this is merged the CI configuration for this project will be able to build the various different Docker images from a single Dockerfile.

Using an ARG directive means that we can parameterise the Docker images we are building by passing `--build-arg SWIFT_...=...` to the "docker build" command.

ARG values are still exposed to RUN commands like ENV, but are not persisted in the final image.

Example usage:

```
docker build \
-t swift:2017-01-28-a \
--build-arg=SWIFT_BRANCH=development \
--build-arg SWIFT_VERSION=swift-DEVELOPMENT-SNAPSHOT-2017-01-28-a \
.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,10 @@ ARG SWIFT_PLATFORM=ubuntu16.04
 ARG SWIFT_BRANCH=swift-3.0.2-release 
 ARG SWIFT_VERSION=swift-3.0.2-RELEASE
 
+ENV SWIFT_PLATFORM=$SWIFT_PLATFORM \
+    SWIFT_BRANCH=$SWIFT_BRANCH
+    SWIFT_VERSION=$SWIFT_VERSION
+
 # Download GPG keys, signature and Swift package, then unpack and cleanup
 RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" | tr -d .)/$SWIFT_VERSION/$SWIFT_VERSION-$SWIFT_PLATFORM.tar.gz \
     && curl -fSsL $SWIFT_URL -o swift.tar.gz \
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,10 +21,9 @@ RUN apt-get -q update && \
     && rm -r /var/lib/apt/lists/*
 
 # Everything up to here should cache nicely between Swift versions, assuming dev dependencies change little
-ENV SWIFT_BRANCH=swift-3.0.2-release \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -21,10 +21,9 @@ RUN apt-get -q update && \
     && rm -r /var/lib/apt/lists/*
 
 # Everything up to here should cache nicely between Swift versions, assuming dev dependencies change little
-ENV SWIFT_BRANCH=swift-3.0.2-release \
-    SWIFT_VERSION=swift-3.0.2-RELEASE \
-    SWIFT_PLATFORM=ubuntu16.04 \
-    PATH=/usr/bin:$PATH
```
swizzlr, line 27: Honestly haven't checked, but are we good with dropping this PATH variable?

swizzlr (commented):
THIS IS AWESOME!

We'll be dropping the env vars. Is it desirable that the swift version is in the environment? Probably not – you can just ask the binary.

I like this.

Dockerfile
aidansteele: I'm not sure why it was there to begin with - maybe a relic of earlier iterations of this file. `/usr/bin` is already on the PATH from the Ubuntu base image:

    docker run -it ubuntu:16.04 env | grep PATH
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

aidansteele:
@swizzlr I don't really mind either way. I'm erring towards removing them - because like you said, they can be determined from `swift` itself.

aidansteele:
Also I don't really understand the CI failure, but it sounds like it's an unrelated config issue?

> failed to obtain yml from provider

swizzlr:
LGTM. Before we bring this into official-land, the only person I can think to check with is @tianon – is there anything bad about using ARG over ENV?

Ignore the CI – @Khalian set it up, he may know.

Khalian:
Odd. This typically happens when shippable cannot find its yml file. But your commit does not change the shippable.yml file in any way. 

I have opened up a issue with them https://github.com/Shippable/support/issues/3348. 

@swizzlr Would you mind restarting the build once. I don't think I have the access permissions to do it. @aidansteele I guess this PR can be merged if you verified that the build process went through. That should unblock you.

tianon:
@swizzlr sorry for being so laggy :disappointed: -- `ARG` should be fine for the official images :+1:

PR #62, opened 2017-02-18
<Description> : Adding support for libssl-dev for projects that work with the OpenSSL bridging header.

Khalian:
<Type> : fix/config

Kitura was failing its build (I was trying to build Kitura in the container in a hackathon). 

note: you may be able to install openssl using your system-packager:

    apt-get install openssl libssl-dev

    Compile Swift Module 'Socket' (3 sources)
    Compile Swift Module 'LoggerAPI' (1 sources)
    Compile Swift Module 'KituraTemplateEngine' (1 sources)
    Compile Swift Module 'SwiftyJSON' (2 sources)
    Compile Swift Module 'SSLService' (1 sources)
    /Kitura/Packages/OpenSSL-0.3.1/module.modulemap:18:15: error: header    '/usr/include/openssl/conf.h' not found
       header "/usr/include/openssl/conf.h"
              ^
    /Kitura/Packages/SSLService-0.12.21/Sources/SSLService.swift:26:9: error: could not build    Objective-C module 'OpenSSL'
        import OpenSSL

After the fix

    Compile Swift Module 'Socket' (3 sources)
    Compile Swift Module 'LoggerAPI' (1 sources) 
    Compile Swift Module 'KituraTemplateEngine' (1 sources)
    Compile Swift Module 'SwiftyJSON' (2 sources)
    Compile Swift Module 'SSLService' (1 sources)
    Compile CHTTPParser http_parser.c
    Compile CHTTPParser utils.c
    Linking CHTTPParser
    Compile Swift Module 'KituraNet' (35 sources)
    Compile Swift Module 'Kitura' (42 sources)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN apt-get -q update && \
     python2.7 \
     python2.7-dev \
     libicu-dev \
+    libssl-dev \
     rsync \
     libxml2 \
     git \


Khalian:
Special shout out to @acbodine for helping me out on this.

swizzlr:
Yeah, so I agree with this in principle, however I'm not sure how I should feel about it, since adding openssl is kind of... should we? It's image bloat; then again, Kitura should be able to cope without it.

Khalian:
The merge commit is unnecessary. A rebase would be just fine. 

Regarding image bloat, the Kitura folks wanted to move to the official docker image. If we support consumers out of the box, the base image has more utility. Otherwise consumers often end up with their own solutions. Also OpenSSL is used by quite a few projects, so overall I think its kind of OK to put it in.

swizzlr:
@Khalian It's also likely that this will end up in Swift Server security APIs.

So moved.

PR #69, opened 2017-04-06
<Description> : Removing binaries that were used for installation but not required post installation.

Khalian:
<Type> : features/config

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,5 +56,9 @@ RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" | 
     && tar -xzf swift.tar.gz --directory / --strip-components=1 \
     && rm -r "$GNUPGHOME" swift.tar.gz.sig swift.tar.gz
 
+# Post cleanup for binaries orthogonal to swift runtime, but was used to download and install. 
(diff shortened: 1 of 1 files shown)


Khalian:
@swizzlr Whenever you get the time.

Khalian:
@swizzlr @hamin Hello ?

hamin:
@Khalian so I think that we can probably remove python package with confidence. I don't think we can remove curl since its used by NSURLSession under the covers. At one point i think we just had libcurl instead of curl installed but that stopped working with the swift images and we had to revert to curl. @Khalian if u could investigate if libcurl is sufficient now then we can probbaly just swap curl for libcurl. I'm not a sure about removing git and make either. I might be mistaken but I think the swift package manager does rely on git and perhaps on make for C deps but i could be wrong. @swizzlr thoughts?

Khalian:
@hamin That is very odd. We would need all of curl to work with NSURLSession? Is there a prototype script/bit of code I can try out with just libcurl ?

Not sure I understand git and make. I removed the two post installation. Are there modules other than Swift Package Manager which need it? And if so is there a requirement to keep the Swift Package Manager in the final draft of a build?

Khalian:
@hamin @swizzlr How about now ?

hamin:
@Khalian I think i like it

PR #71, opened 2017-04-25
<Description> : Creating onbuild variant for docker swift.

Khalian:
https://github.com/swiftdocker/docker-swift/issues/54

<Type> : config

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,3 +58,10 @@ RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" | 
 
 # Print Installed Swift Version
 RUN swift --version
+
+WORKDIR /code
+ONBUILD COPY Package.swift /code/
+ONBUILD COPY Package.pins /code/ 
+ONBUILD COPY Sources /code/
+ONBUILD RUN swift build -c release || true 
+CMD ["/bin/bash", "-c", "find .build -executable -type f -not -name '*.so'| swift package describe | grep "Type: executable" | head | bash"]


swizzlr:
Can you rename this to `Dockerfile-onbuild`?

Dockerfile
```diff
@@ -58,3 +58,10 @@ RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" |
 
 # Print Installed Swift Version
 RUN swift --version
+
+WORKDIR /code
+ONBUILD COPY Package.swift /code/
```
swizzlr, line 63: We should add `Package.pins` as well, IMO.

Dockerfile
```diff
@@ -58,3 +58,10 @@ RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" |
...
+ONBUILD RUN swift build -c release
```
swizzlr: If you could make the build configuration an `ARG` that defaults to `release` then it means people can parametrize the docker build to debug as necessary.

swizzlr: Also add `|| true` to the end of this. If you run swift build without Sources it sometimes returns a non-zero exit code, which fails the build. This command is really an attempt at caching dependency build products, but it's not required to succeed.

Dockerfile
```diff
@@ -58,3 +58,10 @@ RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" |
...
+ONBUILD COPY Sources /code/
+ONBUILD RUN swift build -c release
+CMD ["/bin/bash", "-c", "find .build -executable -type f -not -name '*.so' | head | bash"]
```
swizzlr: This is super clever! However I've noticed that some dependencies include executables, so this might find those examples by accident.

Two ideas:
1. Sort by creation time. The last built executable binary is probably(?) the user's own binary.
2. `grep` `swift package describe` for `Type: executable`, since that will list the non-dependency executables.

`swift package describe` can also output in json, but we'd need `jq` installed for that...

swizzlr (changes requested):
Thanks for contributing all of this! Some notes on making it even better – but then let's merge!

Dockerfile
Khalian: Done, although this looks a little non strict. Under what conditions is the non zero exit code returned ?

Dockerfile
Khalian: Done

Khalian:
"Can you rename this to Dockerfile-onbuild?"

Are you referring to the branch name? Would it matter if it goes into after master post merge?

Dockerfile
Khalian: Done. Is it possible to use the unix file command to get the description instead?

Khalian:
@swizzlr Can we merge this now ? Or is there any other change you need ?

Khalian:
@swizzlr Hello ?

Khalian:
@hamin Would there be anything else needed for this ?

hamin:
hmm i'm actually not a 100% sure about this. @swizzlr thoughts?

Khalian:
@swizzlr @hamin hello ?

andradei:
I'm waiting on this little change as well. Anyone with rights willing to merge this?

hamin:
sorry for the late response @Khalian @andradei , so @Khalian I think what @swizzlr was referring to create a new onbuild variant under the same branch. @Khalian I think probably best would be to create a new `Dockerfile` under `onbuild/Dockerfile` path in the repo. That way I can create a `tag` on Docker hub for `-onbuild` variants. @Khalian since you did the work on this before I don't want that to go to waste, you wanna go ahead and make that change? I'll merge it soon after

Khalian:
@hamin I do not like the idea of two separate Dockerfiles being on the File System on the same time. That is a maintenance headache. I do not even like the idea of two branches, its only slightly better than two Dockerfiles at the same time.

tianon:
https://github.com/docker-library/official-images/issues/2076 is probably of interest -- for all but `sentry`, `onbuild` variants of the official images are officially deprecated (and have been since May), and new ones will not be accepted to the official images without some strong convincing of the benefits (as in the case of the currently-only exception, `sentry`).

swizzlr:
Based on @tianon's comment I'm closing this for now. If someone wants to create an `onbuild` variant based on our image, they can knock themselves out, but implicit behaviors can be a real burden.

Issue #70, opened 2017-04-21
/usr/lib/swift/CoreFoundation not readable for non-root users


helje5:
When running as a regular user, swiftc can't compile some code because the files in `/usr/lib/swift/CoreFoundation` are not readable:

```shell
swift@50cd436932d2:~$ ls -la /usr/lib/swift/CoreFoundation/|head
total 1108
drwxr-xr-x 2 root root  4096 Mar 28 07:02 .
drwxr-xr-x 9 root root  4096 Mar 28 07:02 ..
-rw-r----- 1 root root 31568 Mar 28 07:02 CFArray.h
-rw-r----- 1 root root   788 Mar 28 07:02 CFAsmMacros.h
-rw-r----- 1 root root 11948 Mar 28 07:02 CFAttributedString.h
```

That issue seems to be specific to CoreFoundation, the other directories seem to come with proper permissions.

hamin:
hmm @swizzlr thoughts on this?

swizzlr:
This sounds like an issue with the upstream package, especially if the other directories are fine. We're literally just untarring it. My recommendation would be to figure out what's going wrong with the CI and fix it. In the interim, `RUN chmod...` would do the trick.

I'm curious @helje5, what's the use case for non-root usage? I've never encountered that use case in docker before.

helje5:
Yes, the `RUN chmod` is what I do now as a workaround ([mod_swift-demo.dockerfile](https://github.com/AlwaysRightInstitute/mod_swift/blob/develop/.docker/mod_swift-demo.dockerfile#L36)). Would be cool if both tarball and Docker would be correct out of the box (that is even more important for the tarball ...)

So you are saying the tarball is b0rked. Do you want me to file a Swift issue or are you going to do this?

P.S.: I don't understand the question about the non-root usage. Any leaf container should not run as root for so many reasons. Why would you ever want to give a user process write access to something in /usr, that is just asking for trouble ... Note: I'm not suggesting that this (base) image needs to be setup as non-root. But this is off-topic, if you want to discuss, feel free to send me a message :-) I also still think that it is highly undesirable to have a compiler setup within a deployment container (I'd prefer  separate swift-runtime, swift-compile and swift-repl-play docker images).

P.S.2: I made a Swift 3.1 docker image for RaspberryPi, if there is any interest to integrate this, I could open another issue for this. (https://hub.docker.com/r/helje5/rpi-swift/)

Khalian:
[...]

PR #74, opened 2017-05-07
Adding permissions for Core Foundation library.

Khalian:


Khalian:
Solution to https://github.com/swiftdocker/docker-swift/issues/70

hamin:
@swizzlr I'm ok with this. Although I myself usually use root for docker we shouldn't necessarily stop ppl for being able to use swift with a non-root users in the docker image. This PR I think solves the issue raised #70

Khalian:
Can we merge this ?

Khalian:
@hamin Can we merge this?

hamin:
@Khalian can u make just one change... instead of adding a new `RUN` statement can you append that to the https://github.com/Khalian/docker-swift/blob/d86c55c3003b30184f0fa0d9792a39f8a43b13d8/Dockerfile#L58 w/ `&&` as the last statement run? Once you go ahead and do that i'll merge. And keep your comment in there if you can...super helpful for short term memories :)

Khalian:
@hamin Done.

PR #76, opened 2017-05-14
Adding support for shared library path lookup.

Khalian:


Khalian:
@hamin Can we merge this?

Khalian:
@hamin Would you mind taking a look at this ?

hamin:
@khalian trying to understand why this was needed. What's the issue that this is trying to solve? Was this introduced in a more recently? Thanks.

Khalian:
@hamin It solves this problem https://github.com/swiftdocker/docker-swift/issues/75

hamin:
@khalian ah awesome, is there an example that I can try with? This looks good though

PR #77, opened 2017-07-13
Replace MAINTAINER instruction

rpitonak:
MAINTAINER instruction is deprecated according to 
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


hamin (approved):

hamin:
thank you @rpitonak

PR #84, opened 2017-09-26
cleanup

lgaches:


hamin:
@lgaches are we sure `python` isn't needed and `libpython` alone is sufficient? /cc @swizzlr

hamin:
@lgaches this is looking good to me, i just tested out the `libpython` thing... i'm gonna wait a lil bit to see if @swizzlr has any comments, then I'll merge and create new tags on our docker hub image too.

hamin (approved):

Dockerfile
```diff
@@ -1,16 +1,17 @@
 FROM ubuntu:16.04
-MAINTAINER Haris Amin <aminharis7@gmail.com>
+LABEL maintainer="Haris Amin <aminharis7@gmail.com>"
+LABEL Description="Docker Container for the Apple's Swift programming language"
 
 # Install related packages and set LLVM 3.6 as the compiler
 RUN apt-get -q update && \
+    apt-get dist-upgrade -y && \
```
tianon: Please do not use `dist-upgrade` in submissions to the official images.

tianon: See https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get, especially:

> You should avoid RUN apt-get upgrade or dist-upgrade, as many of the “essential” packages from the parent images won’t upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, you should contact its maintainers. If you know there’s a particular package, foo, that needs to be updated, use apt-get install -y foo to update automatically.

hamin: @lgaches what do u think? I'm trying to remember how we were doing it before. We'd of course like to be compliant w/ docker official image best practices.

lgaches: @hamin I've updated the pull request to be compliant with docker official image best practices.

tianon: Much better :+1:

hamin (approved):

PR #95, opened 2017-11-10
Use Dockerfile.template as root source of truth.

### LINKED ISSUE #92: Backport master improvements to 3.1 tag
Master's received a number of improvements since 3.1 came out. We should back port these and then PR them into official-images.

swizzlr:
This commit takes the lead from
[docker-node](https://github.com/nodejs/docker-node), separating
supported versions into their own folders, using Dockerfiles generated
from template. This commit closes #92 by backporting Clang 3.8 into 3.1.

I have also removed two soon to expire PGP keys from the file and added the new automation key.

### DIFF
--- a/3.1/Dockerfile
+++ b/3.1/Dockerfile
@@ -1,0 +1,61 @@
+FROM ubuntu:16.04
+LABEL maintainer="Haris Amin <aminharis7@gmail.com>"
+LABEL Description="Docker Container for the Apple's Swift programming language"
+
+# Install related packages and set LLVM 3.8 as the compiler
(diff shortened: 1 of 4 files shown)


hamin:
@swizzlr i wanted to do this forever!!!! redis, ruby, and a bunch of other docker images are doing the same thing! I think i mentioned this in some discussion earlier Thank you!

hamin (approved):
HURRAY!!!!

swizzlr:
I've smoke tested both 3 and 4 with two separate Swift 3 and 4 codebases respectively.

PR #101, opened 2017-11-21
point to official repo on hub

CrowdHailer:
I think this is correct


swizzlr:
@CrowdHailer sorry for the delay, thanks for this!

PR #124, opened 2019-01-08
Add support for Swift 4.2.1 - Ubuntu 18.04

shahmishal:


florianreinhart:
What‘s the reason for setting the time zone to LA? I think it should be UTC, which is the default for Docker containers.

shahmishal:
@florianreinhart good catch, I will remove it.

hamin (approved):
This looks great!

ianpartridge (approved):
Looks good to me.  It would be best if we can sort out the flaky keyserver issues as they affected me when I tried `docker build`ing this to test it out.

PR #134, opened 2019-03-26
Add Swift 5.0 Dockerfiles for Ubuntu 16.04 and Ubuntu 18.04

shahmishal:


shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

ianpartridge:
Hey @shahmishal thanks for this. I think we can probably slim down a bit now that clang is shipped with the release https://forums.swift.org/t/which-clang-package-should-we-install/20542/14?u=ianpartridge

Give me a day or so to check and I'll see what I can come up with 🙂

shahmishal:
@ianpartridge  If this is going to take 1-2 day, should we release the Docker image with current Dockerfiles? We can release new updated image with slim down version.

ianpartridge:
For 16.04 this set of packages works for me and reduces the image by ~250MB:

```
RUN apt-get -q update && \
    apt-get -q install -y \
    libatomic1 \
    libedit2 \
    libsqlite3-0 \
    libcurl3 \
    libxml2 \
    libbsd0 \
    libc6-dev \
    binutils \
    libgcc-5-dev \
    libstdc++-5-dev \
    libpython2.7 \
    curl \
    tzdata \
    git \
    pkg-config \
    && rm -r /var/lib/apt/lists/*
```

flovilmart:
Any particular reason for no using

```
ARG UBUNTU_VERSION=16.04
FROM ubuntu:${UBUNTU_VERSION}
```

This allow build with both versions based on an external ARG.

ianpartridge:
Different versions of Ubuntu require different system packages to be installed, so this won't work unfortunately.

ianpartridge:
I think the best way forward is to merge this PR as-is, then iterate on the `Dockerfile`s until we are happy with them. Once we're agreed, we should then update Docker Hub accordingly.

hamin:
This looks great!

hamin (approved):

stephencelis:
Just want to cross-link the issue I opened here: https://bugs.swift.org/browse/SR-10178

It'd be really swell if future Swift releases were coordinated with swift-docker releases so that they don't lag behind. It'd be even better if we could get some automation around swift-docker supporting all of the snapshots that are released on swift.org

ianpartridge:
They're always going to lag behind a bit, because they're gated by approval from the Docker official images team.

Swift 5 had some late landing changes like https://github.com/apple/swift/pull/22273 which will really improve the images when they come out. It's worth taking a little time to get it right.

stephencelis:
@ianpartridge Images can continue to improve after release, though, as you mentioned before. I'm not sure why Apple's side of things can't be synchronized.

The most painful part, though, is Apple not providing a means of testing snapshots with Docker. So instead we're left with completely custom Docker setups or we use https://hub.docker.com/r/norionomura/swift instead.

shahmishal:
PR created for https://hub.docker.com/_/swift 

https://github.com/docker-library/official-images/pull/5613

PR #135, opened 2019-03-29
5.0: optimise installed packages

ianpartridge:
The current Swift 5.0 `Dockerfile`s are a continuation of the images shipped in earlier releases. 

However, there have been some significant changes to the Linux toolchains in Swift 5.0 which mean we have the opportunity to optimise the system packages that are installed in the images. This will significantly reduce the size of the images, benefitting all users.

Specifically, the following major changes occurred in Swift 5:

- Foundation no longer depends on the system ICU but ships its own
- The compiler no longer depends on a system `clang` but ships its own

In addition, we can make a number of other improvements to the images by removing unnecessary packages and only installing those which are genuinely required to build and run Swift programs.

I have spent time [working out the actual requirements of the toolchain](https://forums.swift.org/t/which-clang-package-should-we-install/20542/20). I think the following packages are required:

Runtime requirements:
- `libatomic1` - all Swift programs link this
- `libbsd0` - all Swift programs link this (for `libdispatch`)
- `libcurl4` - Foundation links this
- `libxml2` - Foundation links this
- `tzdata` - Foundation

Build requirements:
- `libc6-dev` - Compiler
- `binutils` - for `gold` linker
- `libgcc-5-dev` - Compiler
- `libstdc++-5-dev` - Compiler/stdlib
- `libedit2` - Package manager
- `libsqlite3-0` - Package manager
- `git` - Package manager
- `pkg-config` - Package manager
- `libpython2.7` - REPL

Making these changes has the following effect to image size:

| Image | Before | After |
| - | - | - |
| 16.04 | 1.69GB | 1.37GB |
| 18.04 | 1.81GB | 1.37GB |

### DIFF
--- a/5.0/ubuntu/16.04/Dockerfile
+++ b/5.0/ubuntu/16.04/Dockerfile
@@ -2,24 +2,22 @@ FROM ubuntu:16.04
 LABEL maintainer="Swift Infrastructure <swift-infrastructure@swift.org>"
 LABEL Description="Docker Container for the Swift programming language"
 
-# Install related packages and set LLVM 3.8 as the compiler
 RUN apt-get -q update && \
     apt-get -q install -y \
-    make \
-    libc6-dev \
-    clang-3.8 \
-    curl \
-    libedit-dev \
-    libpython2.7 \
-    libicu-dev \
-    libssl-dev \
+    libatomic1 \
+    libbsd0 \
+    libcurl3 \
     libxml2 \
+    libedit2 \
+    libsqlite3-0 \
+    libc6-dev \
+    binutils \
+    libgcc-5-dev \
+    libstdc++-5-dev \
+    libpython2.7 \
     tzdata \
     git \
-    libcurl4-openssl-dev \
     pkg-config \
-    && update-alternatives --quiet --install /usr/bin/clang clang /usr/bin/clang-3.8 100 \
-    && update-alternatives --quiet --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.8 100 \
     && rm -r /var/lib/apt/lists/*    
 
 # Everything up to here should cache nicely between Swift versions, assuming dev dependencies change little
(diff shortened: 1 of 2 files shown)


ianpartridge:
@hamin @swizzlr @shahmishal @weissi @tomerd @tanner0101 please review.

weissi (approved):
thanks! lgtm.

norio-nomura:
As far as I tested, produced binary depends on `libicu*` that installed by `apt-get`
```terminal.sh-session
$ ldd ./.build/x86_64-unknown-linux/release/sourcekitten|grep -i libicu
	libicui18nswift.so.61 => /usr/lib/swift/linux/libicui18nswift.so.61 (0x00007ff313105000)
	libicuucswift.so.61 => /usr/lib/swift/linux/libicuucswift.so.61 (0x00007ff312d2f000)
	libicudataswift.so.61 => /usr/lib/swift/linux/libicudataswift.so.61 (0x00007ff31118a000)
	libicuuc.so.55 => /usr/lib/x86_64-linux-gnu/libicuuc.so.55 (0x00007ff30efee000)
	libicudata.so.55 => /usr/lib/x86_64-linux-gnu/libicudata.so.55 (0x00007ff30b975000)
```

norio-nomura:
> As far as I tested, produced binary depends on libicu* that installed by apt-get

Sorry, never mind. That is an indirect dependency through `libxml2`.

shahmishal:
@swift-ci test

shahmishal (approved):

ianpartridge:
Thanks @shahmishal. Am I correct that the CI testing currently only tests that the `Dockerfile`s build and produce an image? It would be good to improve that to then use the image to build and run some Swift programs too. I could take a look at that next week? What do you think?

shahmishal:
@ianpartridge Yes, currently we are just building the Docker images part of CI. However, we can easily improve this by adding additional feature to https://github.com/apple/swift-docker/blob/master/ci_test.py

ianpartridge:
Thanks for the merge 🙂

It would be good to get more confirmation that the new `Dockerfile`s work for people. I've done some initial Kitura testing, it would be great if @weissi or @tomerd could confirm they are a good base for SwiftNIO, and @tanner0101 for Vapor.

PR #149, opened 2019-10-14
Change description label to be lowercase

slashmo:
I know this is pretty nit-picky but when using the Swift images I noticed that they all use the `Description` label instead of `description`. In my opinion this should be changed to go along with [the convention](https://docs.docker.com/engine/reference/builder/#label) and with the `maintainer` label used in the Swift images.

### DIFF
--- a/3.1/Dockerfile
+++ b/3.1/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:16.04
 LABEL maintainer="Haris Amin <aminharis7@gmail.com>"
-LABEL Description="Docker Container for the Apple's Swift programming language"
+LABEL description="Docker Container for the Apple's Swift programming language"
 
 # Install related packages and set LLVM 3.8 as the compiler
 RUN apt-get -q update && \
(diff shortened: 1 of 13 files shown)


ianpartridge:
I'm very happy to make this change in future, from Swift 5.2, but I don't think we should change the labels of images that have already been published.  People could be relying on them:

```
$ docker inspect -f '{{index .Config.Labels "Description"}}' swift:5.1
Docker Container for the Swift programming language
$ docker inspect -f '{{index .Config.Labels "description"}}' swift:5.1

$
```

slashmo:
@ianpartridge Yeah, that makes sense. Sounds good! Should I open a separate issue to note this down for future images?

ianpartridge:
I think we can keep this one open? It'll be a while until we create the Swift 5.2 Docker images.

Thanks for opening this!

slashmo:
@shahmishal The nightly builds are a great addition 👍 Could we use this as an opportunity to start switching to lowercase `description` labels as described above?

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create the pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #194, opened 2020-08-10
Update the Dockerfile from using apt-get update and yum update to mee…

shahmishal:
…t guidelines by hub.docker.com

https://github.com/docker-library/official-images/pull/8509#issuecomment-670627074

> I'm not sure how we missed it previously, but yum -y update shouldn't be used in Dockerfiles -- we recommend instead asking the base image maintainers to update their images if there are packages in the base image that need to be updated (both because often the packages from the base image will fail to update, but more importantly because it severely bloats the size of the resulting images by including many of the same files twice).

>(This is discussed in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get, https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves, and somewhat in https://github.com/docker-library/official-images#library-definition-files.)

### DIFF
--- a/5.2/amazonlinux/2/Dockerfile
+++ b/5.2/amazonlinux/2/Dockerfile
@@ -2,7 +2,7 @@ FROM amazonlinux:2
 LABEL maintainer="Swift Infrastructure <swift-infrastructure@swift.org>"
 LABEL description="Docker Container for the Swift programming language"
 
-RUN yum -y update && yum -y install \
+RUN yum -y install \
   binutils \
   gcc \
   git \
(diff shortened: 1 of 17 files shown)


shahmishal:
@swift-ci test

drexin (changes requested):
The `apt-get update` is required here, because it updates the package lists. The linked article talks about `apt-get upgrade`, which would upgrade the installed packages (which is what `yum update` also does). So removing `yum update` is correct, removing `apt-get update` is not.

shahmishal:
@drexin Thanks I will revert the apt-get update

shahmishal:
@swift-ci test

drexin (approved):
Thanks you!

tomerd:
should we do this in the other (nightly, 5.3, etc) docker files too?

tomerd:
cool! what is https://github.com/apple/swift-docker/tree/master/nightly-5.3 used for? seems like it only has ubuntu?

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

PR #158, opened 2020-03-09
optimization debian package manager tweaks

rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

PR #203, opened 2020-10-07
optimization debian package manager tweaks

rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Also , added packages apt-utils ca-certificates

Because build is

1.  Slow because "apt-utils" not installed

2. to avoid build to exits with error without having certificate in wget , curl or even git clone

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/3.1/Dockerfile
+++ b/3.1/Dockerfile
@@ -4,7 +4,9 @@ LABEL Description="Docker Container for the Apple's Swift programming language"
 
 # Install related packages and set LLVM 3.8 as the compiler
 RUN apt-get -q update && \
-    apt-get -q install -y \
+    apt-get --no-install-recommends install -y \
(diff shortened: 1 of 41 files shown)


rajpratik71:
Fresh pull of #158 to main branch after rebase from upstream

albertoblue87-netizen (approved):
Hecho

PR #218, opened 2021-01-27
lldb now requires python 3.6

weissi:
fixes this error

```
# lldb
lldb: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

which is https://bugs.swift.org/browse/SR-14099

### DIFF
--- a/nightly-main/ubuntu/16.04/Dockerfile
+++ b/nightly-main/ubuntu/16.04/Dockerfile
@@ -10,7 +10,7 @@ RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true && ap
     libcurl3 \
     libedit2 \
     libgcc-5-dev \
-    libpython2.7 \
+    libpython3.6 \
     libsqlite3-0 \
     libstdc++-5-dev \
     libxml2 \
--- a/nightly-main/ubuntu/18.04/Dockerfile
+++ b/nightly-main/ubuntu/18.04/Dockerfile
@@ -10,7 +10,7 @@ RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true && ap
     libcurl4 \
     libedit2 \
     libgcc-5-dev \
-    libpython2.7 \
+    libpython3.6 \
     libsqlite3-0 \
     libstdc++-5-dev \
     libxml2 \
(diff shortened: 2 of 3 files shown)


weissi:
CC @JDevlieghere @shahmishal

JDevlieghere (approved):

weissi:
@swift-ci test

JDevlieghere:
This LGTM, although I wonder if `python3-dev` would be even better.

weissi:
@JDevlieghere I just checked that just `libpython3.6` is good enough to debug. And ideally, these docker images would be as small as possible and therefore I'd say let's go without the `-dev`. If people really want to compile python modules then I think it's fair that they `apt-get install libpython3.6-dev`. WDYT?

JDevlieghere:
> @JDevlieghere I just checked that just `libpython3.6` is good enough to debug. And ideally, these docker images would be as small as possible and therefore I'd say let's go without the `-dev`. If people really want to compile python modules then I think it's fair that they `apt-get install libpython3.6-dev`. WDYT?

Cool, no need to bloat the image if we don't need to. Thanks for checking.

shahmishal (approved):

weissi:
@shahmishal my PR failed because of

```
22:24:13 Error response from daemon: conflict: unable to delete 300e315adb2f (must be forced) - image is referenced in multiple repositories
22:24:13 Error response from daemon: conflict: unable to delete 300e315adb2f (must be forced) - image is referenced in multiple repositories
```

I suspect that's got nothing to do with my PR. I'll retry

weissi:
@swift-ci test

JDevlieghere:
@shahmishal Do you know what the issue is here?

shahmishal:
```
gpg: keyserver receive failed: Cannot assign requested address
```

shahmishal:
Going to merge this

shahmishal (approved):

shahmishal:
@weissi Sorry, I had to revert this PR. 

Ubuntu 16.04:
```
[91mE: Unable to locate package libpython3.6
E: Couldn't find any package by glob 'libpython3.6'
E: Couldn't find any package by regex 'libpython3.6'
```

weissi:
@shahmishal om my, so sorry, I thought I tested that version too but I must have missed at least one. Fixing now

PR #251, opened 2021-10-13
Add support for building Swift-DocC-Render in Swift CI

ethan-kusters:
Bug/issue #, if applicable: rdar://81634420

## Summary

Adds Node 14 as a dependency in the Docker containers used by Swift CI in order to support building Swift-DocC-Render.

Note that this PR does **not** add Node as a dependency in the Docker containers distributed by Swift.org for general use. Node 14 is a build-time dependency for Swift-DocC-Render and not a run-time dependency so it is only required for the specialized Docker containers used by Swift CI for toolchain builds and PR testing.

### DIFF
--- a/ci_test.py
+++ b/ci_test.py
@@ -20,14 +20,14 @@ import sys
 import os
 
 
-def run_command(cmd, log_file=None):
+def run_command(cmd, log_file=None, cwd=None):
     print("Running: {}".format(cmd))
     sys.stdout.flush()
     if log_file:
         file = open(log_file, "w")
-        p = subprocess.Popen(cmd, shell=True, stdout=file, stderr=file)
+        p = subprocess.Popen(cmd, shell=True, stdout=file, stderr=file, cwd=cwd)
     else:
-        p = subprocess.Popen(cmd, shell=True)
+        p = subprocess.Popen(cmd, shell=True, cwd=cwd)
       
     (output, err) = p.communicate()
     return p.wait()
@@ -45,8 +45,7 @@ def get_dockerfiles():
         filename = file_info['filename']
         print(filename)
         if "Dockerfile" in filename:
-            file_dir = filename.replace("Dockerfile", "")
-            dockerfiles.append(file_dir)
+            dockerfiles.append(filename)
     return dockerfiles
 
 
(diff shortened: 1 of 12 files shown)


shahmishal (approved):

shahmishal:
@swift-ci test

tomerd:
this is pretty substantial dependency to add to the toolchain build. how hard is this dependency and what alternatives have been considered? additionally, how is this handled on other platforms?

shahmishal:
@swift-ci test

shahmishal:
@swift-ci test

tkremenek:
> this is pretty substantial dependency to add to the toolchain build. how hard is this dependency and what alternatives have been considered? additionally, how is this handled on other platforms?

I agree this is worth a bit more discussion to see if there are more options to explore before this is merged as is.

ethan-kusters:
> this is pretty substantial dependency to add to the toolchain build.

Hi @tomerd! I agree this is a significant dependency introduction. We agree that requiring Node.js here is sub-optimal and we are working towards removing this dependency entirely.

In the mean-time, we have done our best to mitigate the impact it will have on toolchain builders. You can see some detail about this in @d-ronnqvist's PR (https://github.com/apple/swift/pull/39723). The core of it though is that Node.js is a build-time (not run-time) dependency for Swift-DocC-Render (not Swift-DocC itself), so we are not proposing requiring it generally. 

General users of the Swift toolchains from Swift.org will never need Node.js as they will have access to a pre-built version of Swift-DocC-Render. 

Toolchain builders will only need Node.js if they pass the `--swiftdoccrender` flag to the toolchain build script and explicitly opt-in to building the renderer. We understand most people building toolchains won’t want to install Node on their build machines by default, so the default toolchain scripts will not include DocC Render until we come up with a solution like a tarball they can download. This means other folks building toolchains do NOT need to include Node unless they opt into it. This should keep things pretty simple for most toolchain builders.

> how hard is this dependency and what alternatives have been considered?

This is a hard dependency for building Swift-DocC-Render from source as is required for Swift-CI style toolchain builds from the `main` branch. 

As you suggested offline I think it would be great to also offer a tarball containing a prebuilt version of Swift-DocC-Render on Swift.org. This would allow other toolchain builders who don't want to add a dependency on Node to still fully support Swift-DocC-Render in their toolchains. I would defer to @shahmishal on the feasibility of this.

> additionally, how is this handled on other platforms?

Since this is the first integration of Swift-DocC-Render (and Swift-DocC) with the toolchain we don't currently handle this on any platforms.

With this PR and the one on the Swift toolchain build script (https://github.com/apple/swift/pull/39723) we aim to add support for all toolchains distributed by Swift.org with the notable exception of Windows.

Swift-DocC itself does not yet support Windows but we would like to add support soon. At that time, we would expect to add Node as a dependency to the Windows toolchain builder.

ethan-kusters:
After considering this further, we’ve decided to close this pull request in favor of a solution where we’ll build Swift-DocC-Render in an isolated Docker container and make the build artifact available to Toolchain builders.
This will remove any kind of dependency on Node from the Toolchain build process.
I’ll be updating the existing PR on the Swift build script (https://github.com/apple/swift/pull/39723) to reflect this change.

Thank you for the great feedback here!

PR #274, opened 2022-03-15
update git installation on centos 7 to git 2.x

tomerd:
motivation: centos 7 ships with git 1.x which is too old for the toolchain usage

chages:
* use RHL software collections to install git 2.x
* add symlink since collections install in /opt (TBD better solution)

### DIFF
--- a/nightly-main/centos/7/Dockerfile
+++ b/nightly-main/centos/7/Dockerfile
@@ -2,10 +2,12 @@ FROM centos:7
 LABEL maintainer="Swift Infrastructure <swift-infrastructure@forums.swift.org>"
 LABEL description="Docker Container for the Swift programming language"
 
+RUN yum install -y centos-release-scl-rh
+
 RUN yum install shadow-utils.x86_64 -y \
   binutils \
   gcc \
-  git \
+  rh-git227-git \
   glibc-static \
   libbsd-devel \
   libcurl-devel \
@@ -19,6 +21,8 @@ RUN yum install shadow-utils.x86_64 -y \
   python3 \
   sqlite \
   zlib-devel
+
+RUN ln -s /opt/rh/rh-git227/root/bin/git /usr/bin/git
 
 RUN sed -i -e 's/\*__block/\*__libc_block/g' /usr/include/unistd.h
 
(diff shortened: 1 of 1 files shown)


nightly-main/centos/7/Dockerfile
```diff
@@ -20,6 +22,8 @@ RUN yum install shadow-utils.x86_64 -y \
   sqlite \
   zlib-devel
 
+RUN ln -s /opt/rh/rh-git227/root/bin/git /usr/bin/git
```
tomerd, line 27: I dont like this, and happy to change to a better way to achieve this. the other option of course is adding to scl-rh /opt location to the PATH but that also comes with its own set of issues

tomerd: @drexin @shahmishal wdyt?

tomerd:
@shahmishal @drexin thoughts?

nightly-main/centos/7/Dockerfile
shahmishal: Let's provide a comment explaining why we need git 2.x in the Dockerfile.

tomerd, line 27:
```suggestion
# centos 7 ships with git 1.x which is too old for the toolchain usage
RUN ln -s /opt/rh/rh-git227/root/bin/git /usr/bin/git
```

nightly-main/centos/7/Dockerfile
```diff
@@ -2,10 +2,12 @@ FROM centos:7
 LABEL maintainer="Swift Infrastructure <swift-infrastructure@forums.swift.org>"
 LABEL description="Docker Container for the Swift programming language"
 
+RUN yum install -y centos-release-scl-rh
```
tomerd, line 6:
```suggestion
# centos 7 ships with git 1.x which is too old for the toolchain usage, using RH software collections to install git 2.x
RUN yum install -y centos-release-scl-rh
```

shahmishal (approved):

stevapple:
It turns out that simply symlinking the executable won't work because `git` has runtime dependency on `httpd24`.

I managed to get it work by:
```dockerfile
RUN echo "source /opt/rh/rh-git227/enable" >> /etc/profile.d/sh.local
ENV ENV=/etc/profile.d/sh.local
```
though which the environment variables can be correctly set for both `bash` and `sh`. @tomerd @shahmishal

PR #277, opened 2022-04-11
Introduce a Dockerfile for Windows

compnerd:
This allows one to build a Windows container for Swift.  Although it is
not perfect, it is sufficient for basic tasks.

The Git installation should preferably enable symbolic link support
which although it is marked to do does not seem to be enabled.

The Python installation should preferably set `PYTHONHOME` and
`PYTHONPATH`.  This will become increasingly important as further
refinements to the toolchain start getting merged.

### DIFF
--- a/5.6/windows/20H2/Dockerfile
+++ b/5.6/windows/20H2/Dockerfile
@@ -1,0 +1,49 @@
+# escape=`
+
+FROM mcr.microsoft.com/windows/servercore:20H2 AS windows
+
+LABEL maintainer="Swift Infrastructure <swift-infrastructure@forums.swift.org>"
+LABEL description="Docker Container for the Swift programming language"
+
+ARG GIT=https://github.com/git-for-windows/git/releases/download/v2.35.1.windows.2/Git-2.35.1.2-64-bit.exe
+ARG INSTALLER=https://download.swift.org/swift-5.6.1-release/windows10/swift-5.6.1-RELEASE/swift-5.6.1-RELEASE-windows10.exe
+ARG PYTHON=https://www.python.org/ftp/python/3.10.4/python-3.10.4-amd64.exe
+
+# restore the default Windows shell for correct batch processing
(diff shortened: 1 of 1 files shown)


tomerd:
very interesting! can this be used from any host or only windows one?

compnerd:
I would consult the EULA: https://docs.microsoft.com/en-us/virtualization/windowscontainers/images-eula.  I am not sure how to really interpret this sadly.

5.6/windows/Dockerfile
```diff
@@ -0,0 +1,49 @@
+# escape=`
```
shahmishal: What is this for?

shahmishal:
Overall looks good, can you move this under version dir? `5.6/windows/<windows_version>/Dockerfile` 

 thanks @compnerd!

5.6/windows/Dockerfile
compnerd: Ah, it is needed to get the line continuation to work.  I don't fully understand the _why_, but this is actually a control message - similar to a shebang.

compnerd:
Oh, a Windows version, sure!  Didn't think to do that ... seems kinda silly in hindsight.

shahmishal:
> Oh, a Windows version, sure! Didn't think to do that ... seems kinda silly in hindsight.

Incase in the future, we have multiple version support. For example: https://github.com/apple/swift-docker/tree/main/5.6/ubuntu

compnerd:
Oh, I totally get why, just seems silly that I didn't think to do that, especially when I know that the image actually has a hard version dependency and we would likely want more than one version.

compnerd:
I'm currently using the "friendly" name (20H2).  I suppose that we could switch to the build number, but I don't know what is easier to work with.  It shouldn't be too bad to move between the two, but I think it requires a bit more experience (or perhaps you have that already and can guide me).

compnerd:
@swift-ci please test

PR #302, opened 2022-06-03
Windows: add a toolchain image for 10.0.19044.1706 (21H2)

compnerd:
Add rules for building a docker image for the toolchain build at
10.0.19044.1706 (21H2).  As the base image for the release is not
available, use a slightly older release which should be compatible
to avoid any issues due to kernel mismatches (the newer image will
not run).  This allows building the toolchain in a docker image with
the latest stable release of Windows 10.


compnerd:
Requires apple/swift#59276 for the build script.

shahmishal:
@compnerd What is the status of this PR?

compnerd:
I think it's ready with the associated change.

PR #306, opened 2022-07-28
Fix git usability for CentOS 7 nightly images

stevapple:
Git 2.x introduced in #274 is not enabled correctly, because its networking feature depends on `httpd24` from RH software collections, which is not in the library search path.

This PR attempts to enable the Git installation correctly by using the official `enable` script. This also enhances compatibility and robustness against any possible dependency changes.


stevapple:
cc @shahmishal @tomerd

stevapple:
ping again @shahmishal @tomerd

shahmishal:
@swift-ci test

shahmishal:
Do we need to pull this into 5.7/nightly-5.7 Dockerfiles too?

stevapple:
@shahmishal Git 2.x is not enabled on 5.7 images yet, so is this patch.

PR #341, opened 2023-05-29
Reinstate "Use Swift 5.7 base image for Swift CI jobs to support bootstrapping"

DougGregor:
With https://github.com/apple/swift/pull/66199, we're close(r) to the point where we can build the Swift compiler that integrates Swift code built using the host Swift.


DougGregor:
@swift-ci please test

swift-ci/master/amazon-linux/2/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM amazonlinux:2
+FROM swift:5.7-amazonlinux2
```
tomerd, line 1: should this be pinned to 5.7 or use the latest release?

bnbarham: Pinned makes sense to me given that we support releases up to 12 months back. So 5.7/5.8/.5.9. Though ideally we'd have a build for 5.8 + 5.9 going as well.

swift-ci/master/centos/7/Dockerfile
```diff
@@ -48,6 +48,38 @@ RUN ldconfig
...
+    && mkdir -p /opt/swift \
+    # - Unpack the toolchain, set libs permissions, and clean up.
+    && tar -xzf swift.tar.gz --directory /opt/swift/ --strip-components=1 \
+    && chmod -R o+r /opt/swift/usr/lib/swift \
+    && ln -s /opt/swift/usr/lib/swift /usr/lib/swift \
+    && rm -rf "$GNUPGHOME" swift.tar.gz.sig swift.tar.gz
+
+ENV PATH="${PATH}:/opt/swift/usr/bin"
+
+RUN swift --version
```
tomerd, line 81: this is a surprising change. why do we need to do this? should we fix the docker image instead if this is fixing something?

bnbarham: Ideally *all* our release images would install into `/opt/swift/<version>` instead of `/usr`. But we don't want to break current clients, so for now we're just updating the builder image.

Without this change, Swift uses the toolchain-clang to build and then uses the system C++ stdlib instead of devtoolset as it should. The system stdlib is wildly out of date (unsurprising) and thus doesn't have any new C++ features.

@shahmishal given all the other images are using the toolchain clang, perhaps we should make this `ENV PATH=/opt/swift/usr/bin:${PATH}` instead of appending?

tomerd:
> Ideally all our release images would install into /opt/swift/<version> instead of /usr

this is where I was going with my questions. I think we should fix this going forward

tomerd:
cc @etcwilde @MaxDesiatov

swift-ci/master/centos/7/Dockerfile
shahmishal: > @shahmishal given all the other images are using the toolchain clang, perhaps we should make this ENV PATH=/opt/swift/usr/bin:${PATH} instead of appending?

Yes, that make sense.

shahmishal: > this is a surprising change. why do we need to do this? should we fix the docker image instead if this is fixing something?

We can change this in 5.9 release, if the community is ok with the change. I would like to avoid this type of change in dot release.

etcwilde:
Yep, using `/opt/swift-<version>` means we can install toolchains for multiple versions of swift (important since we don’t have ABI stability guarantees on Linux), and so that we don’t conflict with the distribution toolchains. This is consistent with how redhat and centos bundle additional and newer toolchains. It’s also consistent the UNIX directory hierarchy spec. 
> reserved for the installation of add-on application software packages.

It gives us a lot more flexibility for how we want to distribute things.

tomerd:
fwiw, not all distributions like things in /opt so this may need to be refined per distribution, but 100% with the approach and we should def fix it in the toolchain packaging itself since its more correct

bnbarham:
I've updated to use a downloaded toolchain rather than the base image for all Linux platforms - https://github.com/apple/swift-docker/pull/346. Made a new PR since I wanted to rebase + squash.

PR #346, opened 2023-07-06
Install a Swift 5.8.1 toolchain in all swift-ci images

bnbarham:
Installs a Swift 5.8.1 toolchain into `/opt/swift/<version>` on each Linux platform. Removes clang since the toolchain clang can be used instead.

### DIFF
--- a/swift-ci/master/amazon-linux/2/Dockerfile
+++ b/swift-ci/master/amazon-linux/2/Dockerfile
@@ -46,7 +46,6 @@ RUN mkdir -p /usr/local/lib/python3.7/site-packages/
 
 RUN easy_install-3.7 six
 
-ARG SWIFT_SIGNING_KEY=A62AE125BBBFBB96A6E042EC925CC1CCED3D1561
 ARG SWIFT_PLATFORM=amazonlinux2
 ARG SWIFT_VERSION=5.8.1
 ARG SWIFT_BRANCH=swift-${SWIFT_VERSION}-release
@@ -54,8 +53,7 @@ ARG SWIFT_TAG=swift-${SWIFT_VERSION}-RELEASE
 ARG SWIFT_WEBROOT=https://download.swift.org
 ARG SWIFT_PREFIX=/opt/swift/${SWIFT_VERSION}
 
-ENV SWIFT_SIGNING_KEY=$SWIFT_SIGNING_KEY \
-    SWIFT_PLATFORM=$SWIFT_PLATFORM \
+ENV SWIFT_PLATFORM=$SWIFT_PLATFORM \
     SWIFT_VERSION=$SWIFT_VERSION \
     SWIFT_BRANCH=$SWIFT_BRANCH \
     SWIFT_TAG=$SWIFT_TAG \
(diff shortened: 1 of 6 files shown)


DougGregor (approved):
There's no way to share this code across Dockerfiles?

bnbarham:
@shahmishal has some docker-compose ideas here IIUC.

With the base docker... not really. You could share a common base image, but this is the other way around - ie. we have a bunch of images that we then want to add a toolchain on top of. Could do some generator-like hackery I suppose, but I'm not sure that's really better (eg. could just have a makefile that concats a bunch of files together 😅). Another alternative would be something like edrevo/dockerfile-plus.

shahmishal:
@swift-ci test

swift-ci/master/amazon-linux/2/Dockerfile
```diff
@@ -47,6 +46,51 @@ RUN mkdir -p /usr/local/lib/python3.7/site-packages/
...
+    && gpg --batch --quiet --keyserver keyserver.ubuntu.com --recv-keys "$SWIFT_SIGNING_KEY" \
+    && gpg --batch --verify swift.tar.gz.sig swift.tar.gz \
+    # - Unpack the toolchain, set libs permissions, and clean up.
+    && mkdir -p $SWIFT_PREFIX \
+    && tar -xzf swift.tar.gz --directory $SWIFT_PREFIX --strip-components=1 \
+    && chmod -R o+r $SWIFT_PREFIX/usr/lib/swift \
+    && rm -rf "$GNUPGHOME" swift.tar.gz.sig swift.tar.gz
+
+ENV PATH="${SWIFT_PREFIX}/usr/bin:${PATH}"
+
```
tomerd, line 91: should we use the RPM packages instead?

tomerd:
> There's no way to share this code across Dockerfiles?

some of this can be made into a script that is that can be copied & used in the docker file

bnbarham:
> > There's no way to share this code across Dockerfiles?
> 
> some of this can be made into a script that is that can be copied & used in the docker file

Ah that's a good point. Could do that, the extra curl for Ubuntu is a little annoying but we could install based on a parameter for that script.

tomerd:
> the extra curl for Ubuntu is a little annoying but we could install based on a parameter for that script.

a bit less efficient from docker caching pov but we can install curl on ubuntu before the ADD and RUN commands for the script, then remove after. or even a more lazy approach, we can just include curl on these images in the main dependencies list even if we dont really need it for anything else

swift-ci/master/amazon-linux/2/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM amazonlinux:2
 
 RUN yum install shadow-utils -y
 
-RUN groupadd -g 998 build-user && \
+RUN groupadd -g 1000 build-user && \
```
MaxDesiatov: I don't have any opinion on the group number, just curious why this change was made?

bnbarham: I cannot remember why this was needed. It definitely was for CentOS when using the release image, but I don't know if it is still needed now that we're not. I'm guessing probably not? I don't suppose you remember @shahmishal / @DougGregor?

DougGregor: I have no memory of this place

bnbarham: I've reverted for now. We'll see if there's any errors pretty quick.

MaxDesiatov:
@swift-ci test

swift-ci/master/amazon-linux/2/Dockerfile
MaxDesiatov: We currently mark those as "experimental use only" on swift.org/download. Are we willing to try those out here anyway, or is description on that page outdated?

bnbarham:
@swift-ci test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

PR #350, opened 2023-08-28
[DNM] Check to see if readelf is on UBI

bnbarham:


bnbarham:
@swift-ci please test

PR #352, opened 2023-09-08
Swift 5.9-dev: CentOS7: Symlink libstdc++_nonshared.a into static swift

etcwilde:
The swift compiler and runtime libraries are built against the devtoolset-8 on centOS-7. The devtoolset has additional functionality on top of the base centos-7 libstdc++. When the standard library is built as a dynamic library, the extra functionality is linked into the DSO directly, so the dependency is hidden. When the standard library is built as a static archive, it the extra functionality isn't. This is correct behavior to avoid over-linking in the final executable, but it means that the library needs to be present when building statically and in a place that the compiler will look for resources.

This symlinks the backpack library with the additional functionality, along with the linker script to tell the linker to link it into anything that needs to use libstdc++.so.

rdar://114669418

### DIFF
--- a/nightly-5.9/centos/7/Dockerfile
+++ b/nightly-5.9/centos/7/Dockerfile
@@ -60,6 +60,19 @@ RUN set -e; \
     && chmod -R o+r /usr/lib/swift \
     && rm -rf "$GNUPGHOME" latest_toolchain.tar.gz.sig latest_toolchain.tar.gz
 
+# The devtoolset-8 that the Swift runtime is built against uses new
+# functionality in the backdeploy `libstdc++_noshared.a` in devtoolset-a.
+# A linkerscript in the devtoolset ensures that it is picked up appropriately.
+# When dynamically linking the runtime, this static archive is merged into the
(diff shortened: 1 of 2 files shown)


ktoso (approved):

etcwilde:
@swift-ci please test

tomerd (approved):

tomerd (approved):
we probably need this on 5.8 and main as well

etcwilde:
`No package devtoolset-8 available.`
hmm... well that's interesting. Will have to take a look on Monday.

Yeah, we'll probably need it for main. I tried 5.8 with a little statically-linked hello-world and that seemed to work. I can try something a little more complicated though and see if we need this trick there too.

etcwilde:
Okay, confirmed that statically linking both 5.8 and nightly-5.8 work without these changes. nightly-main and nightly-5.9 needed the change, and the eventual 5.9 will need this symlink dance.

etcwilde:
@swift-ci please test

tomerd:
@shahmishal ptal

shahmishal (approved):

PR #359, opened 2023-10-24
Windows: update the LTSC channel image definition

compnerd:
This brings the image definition inline with the 20H2 one. It addresses feedback from the docker library maintainers.

### DIFF
--- a/5.9/windows/LTSC2022/Dockerfile
+++ b/5.9/windows/LTSC2022/Dockerfile
@@ -1,46 +1,115 @@
 # escape=`
 
-FROM mcr.microsoft.com/windows/servercore:ltsc2022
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

5.9/windows/LTSC2022/Dockerfile
```diff
@@ -1,46 +1,115 @@
 # escape=`
 
-FROM mcr.microsoft.com/windows/servercore:ltsc2022
+FROM mcr.microsoft.com/windows/servercore:ltsc2022-amd64 AS windows
```
shahmishal, line 3: not sure if we need the `-amd64` here.

PR #360, opened 2023-10-27
Windows: address further comments from docker-library

compnerd:
`Start-Process` may silently mishandle error handling even with error handling setup. Explicitly check exit codes. Switch to the Windows base image with Python from Microsoft to help shave off ~1G from the image size and reduce complexity. Unfortunately, this requires a bit more work on our side to setup - we need to setup some environment variables.

Remove the 20H2 image as it was silently failing to install python and there are not preexisting tags for the base image with python.

Tested on Windows 11 Pro 22H2 (22621.2428).


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

shahmishal (approved):

PR #361, opened 2023-11-02
Windows: restore the python installation and change the base image

compnerd:
Restore the python installation and the base image as the base image is a requirement for the docker-library.


compnerd:
@swift-ci please test

shahmishal (approved):

PR #362, opened 2023-11-03
Windows: swap install order for python and VSB

compnerd:
Supposedly the ordering matters and reduces the size by ~900M.


compnerd:
@swift-ci please test

shahmishal (approved):

PR #379, opened 2024-04-07
Correct the padding on the /etc/motd banners

jonEbird:
The prior tab based padding of the /etc/motd banner lended itself to an error in the padding in length depending on the value of the Swift tag contents. The approach of using `printf` is a bit more friendly from a maintenance perspective.

Lastly, there was a bug in how the bash code was included where the value of `$TERM` wasn't be escaped properly so you'd get a hardcoded value vs. the variable `$TERM`.

### DIFF
--- a/nightly-5.10/amazonlinux/2/Dockerfile
+++ b/nightly-5.10/amazonlinux/2/Dockerfile
@@ -62,13 +62,14 @@ RUN set -e; \
 # Print Installed Swift Version
 RUN swift --version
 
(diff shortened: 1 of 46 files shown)


shahmishal:
@swift-ci test

shahmishal:
@jonEbird Can we limit this change to active nightly Dockerfiles? 

* main
* 6.0
* 5.10

jonEbird:
> @jonEbird Can we limit this change to active nightly Dockerfiles?
> 
> * main
> * 6.0
> * 5.10

Sure thing. I've updated the commit to only include those.

shahmishal:
@swift-ci test

PR #386, opened 2024-05-25
Install cmake and git test dependencies

etcwilde:
You can build a complete toolchain without git or CMake installed, but it won't test. build-script builds its own copy of CMake, but doesn't add it to its path for testing purposes, so it'll fail to pass its test suite if you don't pre-install a CMake. The tests also need a pre-installed git to check branch state.

rdar://128502727


shahmishal (approved):

shahmishal:
@swift-ci test

PR #397, opened 2024-06-21
Add Swift PATH to /etc/profile for Debian 12

futurejones:
CI builds for Debian 12 are consistently failing at the same point.

The issue is being caused by being unable to find the host Swift install for bootstrapping.
See build logs -
`-- Warning: {} Host toolchain could not locate a compiler to build swift-driver. `

This problem seems to be unique to Debian Docker containers being started via `ssh` e.g. a Jenkins pipeline build.
When the container is started via `ssh` the saved `env PATH` is overwritten.

The solution is to add the PATH to `/etc/profile`

### DIFF
--- a/swift-ci/master/debian/12/Dockerfile
+++ b/swift-ci/master/debian/12/Dockerfile
@@ -78,6 +78,8 @@ RUN set -e; \
 
 ENV PATH="${SWIFT_PREFIX}/usr/bin:${PATH}"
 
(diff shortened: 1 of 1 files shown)


futurejones:
@shahmishal I have run a build on [Swift Community CI Server](https://ci.swiftlang.xyz/job/swift-main-debian-bookworm/440/) to verify the fix.

shahmishal:
@swift-ci test

shahmishal (approved):
Thanks

PR #401, opened 2024-08-01
Use just built CMake for Static SDK

shahmishal:


shahmishal:
@swift-ci test

swift-ci/sdks/static-linux/scripts/build.sh
```diff
@@ -650,6 +650,7 @@ EOF
     # Find some tools
     swiftc=$(find ${build_dir}/swift -name 'swiftc' | grep -v bootstrapping)
     lld=$(find ${build_dir}/swift -name 'ld.lld')
+    cmake=$(find ${build_dir}/swift -type f -name 'cmake')
```
etcwilde: It looks like this is defined after the uses above.
Since we know that we'll need to pre-build CMake, perhaps we should just pre-build it directly then add it to the PATH so that build-script doesn't need to rebuild it, and the others will all pick it up too.

al45tair (approved):

shahmishal:
@swift-ci test

etcwilde (approved):

PR #416, opened 2024-09-17
[static sdk] Update build.sh to include swift-collections deps

shahmishal:


shahmishal:
cc: @etcwilde @al45tair

al45tair (approved):

etcwilde:
I don't think this will fix the missing archive issue, maybe just speed up CI since it won't try to download the sources again here: https://github.com/apple/swift-foundation/blob/ce79f53ac40f5d4663e2c34b61ff81df911028e8/CMakeLists.txt#L74-L83

We need to actually have it install the static archive and swiftmodule associated with collections.

PR #402, opened 2024-08-07
Install libc++ package on CI images of Ubuntu 22.04

egorzhdan:
This adds a `libc++-15-dev` package to the list of packages installed on a CI image of Ubuntu 22.04.

https://packages.ubuntu.com/jammy/libc++-15-dev

Swift/C++ interoperability will soon allow compiling with a C++ standard library that is not the default on the particular platform, e.g. with libc++ on Linux. Having the libc++ package installed in Swift CI allows us to run compiler tests for this new functionality.

rdar://118357548 / https://github.com/swiftlang/swift/issues/69825

PR #420, opened 2024-09-20
Install libc++ package on CI images of Ubuntu 24.04

egorzhdan:
This adds a `libc++-18-dev` package to the list of packages installed on a CI image of Ubuntu 24.04.

Swift is now capable of compiling code that uses C++ interop with libc++ on Linux. Having the libc++ package installed provides us with CI coverage of the feature.

This is similar to https://github.com/swiftlang/swift-docker/pull/402.

### DIFF
--- a/swift-ci/main/ubuntu/24.04/Dockerfile
+++ b/swift-ci/main/ubuntu/24.04/Dockerfile
@@ -11,6 +11,8 @@ RUN apt-get -y update && apt-get -y install \
   cmake                 \
   git                   \
   icu-devtools          \
+  libc++-18-dev         \
+  libc++abi-18-dev      \
   libcurl4-openssl-dev  \
   libedit-dev           \
   libicu-dev            \


finagolfin:
This will probably fail because of swiftlang/swift#76325, as I just tried building those new libc++ tests you added natively on Android with the libc++ 18 from Android NDK 27b and both failed.

You may have to stick with libc++ 16 until that problem is fixed.

egorzhdan:
I haven't been able to reproduce the libc++ issue you're seeing. I tried both macOS and Linux. I suspect this is something Android-specific.

finagolfin:
OK, good that it works for you on linux/mac, but I don't see how it could be Android-specific when it just uses the same libc++ and accompanying C++ module map pretty much, while the failing header is in libc++ itself.

I need to look into those C++ Interop tests that also fail on Android with this libc++ 17 re-org, so I will do some digging.

egorzhdan:
@swift-ci please test

egorzhdan:
@finagolfin we aren't _building_ the compiler with libc++ on Linux, we're only _testing_ it with libc++, so there isn't a risk of the failure you were seeing from this patch.

finagolfin:
Sure, but my subsequent point in this thread stands that these new libc++ tests of yours in the compiler validation suite failed natively on Android with its libc++ 18, which isn't very different from the libc++ package on linux that you're adding here.

I'm not saying don't merge this, as it's certainly possible I'm seeing failures specific to Android somehow, just warning you that's what I saw on another non-CI platform and I'm trying to figure out what might be going wrong.

If you're confident this will work well for linux, don't worry about Android. I will look into those Android failures later and report back.

al45tair (approved):

PR #428, opened 2024-11-08
Add missing system packages needed for C++ on Debian 12, Fedora 39

cmcgee1024:
While compiling swiftly on these new platforms there were missing headers and archives for Debian 12 and Fedora 39. Add these packages so that if the container is used to compile and link C++ code the necessary headers and archives are present.


shahmishal:
Thanks for creating the PR. Can we order the packages alphabetically?

PR #430, opened 2024-11-12
Use fixed version build tools link

plemarquand:
The current link to the VS build tools points to an evergreen bootstrapper. This means that when a new v17 buildtools.exe is available the link is updated with the new version.

Since we're verifying that the downloaded buildtools.exe file hash matches a known value the windows Docker setup breaks each time a new buildtools is released.

Instead, use a fixed version permalink that will not change. The specific version used is version `17.12.0`, build `17.12.35506.116`.

See: https://learn.microsoft.com/en-gb/visualstudio/releases/2022/release-history#fixed-version-bootstrappers

### DIFF
--- a/5.10/windows/1809/Dockerfile
+++ b/5.10/windows/1809/Dockerfile
@@ -89,8 +89,8 @@ RUN Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:PY39});              
     Remove-Item -ErrorAction SilentlyContinue -Force -Recurse ${env:TEMP}\*
 
 # Install Visual Studio Build Tools
-ARG VSB=https://aka.ms/vs/17/release/vs_buildtools.exe
-ARG VSB_SHA256=D4E08524CB0E5BD061A24F507928D1CFB91DCE192C5E12ED964B8343FC4CDEDD
+ARG VSB=https://download.visualstudio.microsoft.com/download/pr/5536698c-711c-4834-876f-2817d31a2ef2/c792bdb0fd46155de19955269cac85d52c4c63c23db2cf43d96b9390146f9390/vs_BuildTools.exe
+ARG VSB_SHA256=C792BDB0FD46155DE19955269CAC85D52C4C63C23DB2CF43D96B9390146F9390
 RUN Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:VSB});               \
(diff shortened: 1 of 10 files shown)


shahmishal:
cc: @compnerd

shahmishal (approved):

compnerd:
I do like the idea that we are pinned for the docker image creation. However, this also means that we need someone to continually update the compiler. Right now, the docker image creation process would inform us that the toolchain has been updated, after this change, we would need to have some sort of process to ensure that we do get periodic updates for the toolchain.

plemarquand:
Is it important that we stay on the latest build tools? I had assumed we'd only really need to update once we need a new version of a build dependency. 

One big issue with the way things stand right now is that no one can build on top of these images to make their own. At any given time someone doing a `docker build` with a FROM swift-docker:whatever-windows-image may or may not have a working base image to use, and its out of their control until we fix the SHA.

compnerd:
For the CI images, I think that it is quite important that we stay on the latest build tools. The build tools includes the C/C++ runtime headers and we need/want those to be up-to-date.

For user images, I think that updating at the time of release is not too terrible (I wouldn't call it ideal, but that is something that we could be a bit flexible on).

compnerd:
Also, perhaps this is a good time to update the images to 17.12.0 😆.

plemarquand:
You're certainly right, I've updated to the latest version 17.12.0.

compnerd:
For the fixed version docker images, I think that the pinned version makes sense. For the nightly and swift-ci images, I'm less convinced.

plemarquand:
Is the guidance to just update the SHA for the nightly and swift-ci images? How will we know the SHA needs to be updated when the build tools change?

compnerd:
Well, the nightly image creation should hopefully be running nightly and tell us that the build has failed. Likewise, for the swift-ci images, we should be updating that semi-regularly and should be able to tell when the deployment fails.

plemarquand:
@compnerd I've reverted the nightly and ci images to use the evergreen link, updating the SHA to match the latest version of the file. 

I'm still a little confused as to how/why this SHA is out of date if things are supposed to be failing in nightly builds whenever they update the build tools.

compnerd (approved):

plemarquand:
@shahmishal Is this ready to merge?

compnerd:
Yeap, this is ready to go!

PR #442, opened 2025-02-19
Windows: tweak LTSC2022 nightly image creation

compnerd:
- Use HTTPS for the latest download info.
- Immediately convert from JSON to help make failures more understandable.
- Perform the JSON conversion once to avoid repeated parsing.

### DIFF
--- a/nightly-main/windows/LTSC2022/Dockerfile
+++ b/nightly-main/windows/LTSC2022/Dockerfile
@@ -127,11 +127,11 @@ RUN Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:VSB});               
     Remove-Item -ErrorAction SilentlyContinue -Force -Recurse ${env:TEMP}\*
 
 # Install Swift toolchain.
-ARG SWIFT_RELEASE_METADATA=http://download.swift.org/development/windows10/latest-build.json
-RUN $env:Release = curl.exe -sL ${env:SWIFT_RELEASE_METADATA};                  \
-    $env:SWIFT_URL = "\"https://download.swift.org/development/windows10/$($($env:Release | ConvertFrom-JSON).dir)/$($($env:Release | ConvertFrom-JSON).download)\"";   \
-    Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:SWIFT_URL});         \
-    Invoke-WebRequest -Uri ${env:SWIFT_URL} -OutFile installer.exe;             \
+ARG SWIFT_RELEASE_METADATA=https://download.swift.org/development/windows10/latest-build.json
+RUN $Release = curl.exe -sL ${env:SWIFT_RELEASE_METADATA} | ConvertFrom-JSON;   \
+    $SWIFT_URL = "\"https://download.swift.org/development/windows10/$($Release.dir)/$($Release.download)\""; \
+    Write-Host -NoNewLine ('Downloading {0} ... ' -f ${SWIFT_URL});             \
+    Invoke-WebRequest -Uri ${SWIFT_URL} -OutFile installer.exe;                 \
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

shahmishal (approved):

weliveindetail:
@compnerd Looks like this change broke the [xInfrastructure docker](https://ci-external.swift.org/view/xInfrastructure/job/publish-nightly-main-docker-images/41/) bot:
```
 ConvertFrom-Json : Cannot bind argument to parameter 'InputObject' because it is null.
At line:1 char:221
  + ... org/development/windows10/$($($env:Release | ConvertFrom-JSON).dir)/$ ...
  +                                                  ~~~~~~~~~~~~~~~~
      + CategoryInfo          : InvalidData: (:) [ConvertFrom-Json], ParameterBindingValidationException
      + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.PowerShell.Commands.ConvertFromJsonCommand
```

Attaching the full log here: [xInfrastructure-docker-41.log](https://github.com/user-attachments/files/18939902/xInfrastructure-docker-41.log)

compnerd:
@weliveindetail that seems odd - this change removes the referenced code 🤔

weliveindetail:
Oh right! Interesting, the checkout was fine and the subsequent commit didn't touch this code. And still, the build was running the old code?

It turns out that CI is running a *different Dockerfile*, which wasn't updated: https://github.com/compnerd/swift-docker/blob/ac4d1fe17ceedb2194ebcd5ac6beef5cbcda8245/nightly-main/windows/1809/Dockerfile#L132

But why does the old code fail now? I compared to the previous (successful) build. The same RUN line worked well there. There was no other change in the meantime. If the URL is not reachable, we should get a 404 as HTML code, and ConvertFrom-JSON would complain with `Invalid JSON primitive`.

Not sure what the issue is. Maybe an internal error in `curl.exe`? I can reproduce the error output locally like this, but we should see error output from `curl.exe` as well and we don't have that in the logs:
```
PS S:\> $env:Release = curl.exe -sL ${env:INVALID}
curl: (2) no URL specified
curl: try 'curl --help' for more information

PS S:\> $env:Release | ConvertFrom-JSON
ConvertFrom-Json : Cannot bind argument to parameter 'InputObject' because it is null.
At line:1 char:16
+ $env:Release | ConvertFrom-JSON
+                ~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [ConvertFrom-Json], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.PowerShell.Commands.ConvertFromJsonCommand      
```

Furthermore, the PR-test for this patch failed, but it seems unrelated: https://ci.swift.org/job/pr-swift-docker/156/

weliveindetail:
The next build failed with the same error: [xInfrastructure-docker-42.log](https://github.com/user-attachments/files/18984244/xInfrastructure-docker-42.log) Only difference is the checkout from https://github.com/apple/swift-continuous-integration-library, that I have no access to:
```diff
--- xInfrastructure-docker-41.log
+++ xInfrastructure-docker-42.log
@@ -14,4 +14,4 @@
  > git config core.sparsecheckout # timeout=10
- > git checkout -f a9bd1bcd3594200f230a5aa70dcb8d423ad7c4ae # timeout=10
-Commit message: "Update macOSNewXcodeJob.groovy"
+ > git checkout -f ff79c48990546dbbfdae5c96eecb60f3ed8cf2d4 # timeout=10
+Commit message: "Add support for build command"
 [Pipeline] Start of Pipeline
```

Assuming for now that the issue originates from there.

PR #450, opened 2025-03-27
Update Git version in some windows Docker images

bkhouri:
Update the Git to v 2.49.0 in the nightly-6.0, nightly-6.1 and nightly-main docker images.

relates to:
- https://github.com/swiftlang/swift-package-manager/issues/8176
- https://github.com/swiftlang/swift-package-manager/issues/8121

PR #452, opened 2025-03-27
Revert "Revert "Update Git version in some windows Docker images (#450)""

bkhouri:
Reverts swiftlang/swift-docker#451, which reverted #450.  However, https://github.com/swiftlang/swift/issues/80356 may be the actual root cause, hence the reason for reverting #451.

The lack of CI pipeline console output created #451 as a speculative revert.


jakepetroules:
It wasn't the git version, this should be good to re-land.

jakepetroules (approved):

PR #454, opened 2025-03-30
Add missing `LD_LIBRARY_PATH` to Dockerfiles for Debian 12/13 and Fedora 39/41

kkebo:
This PR fixes the following issue that you can not run executables built outside the container.

```console
$ cd
$ mkdir test
$ cd test
$ swift package init --type executable
...
$ swift build -c release
...
$ docker container run -it --rm -v $HOME/test:/test swiftlang/swift:nightly-main-debian12
...
build-user@087dcbadce38:~$ /test/.build/release/test
/test/.build/release/test: error while loading shared libraries: libswiftCore.so: cannot open shared object file: No such file or directory
```

The reason why `LD_LIBRARY_PATH` is needed is Dockerfiles for Debian 12 and Fedora39 install libraries under `${SWIFT_PREFIX}/usr/lib` instead of `/usr/lib`.

### DIFF
--- a/nightly-6.1/debian/12/Dockerfile
+++ b/nightly-6.1/debian/12/Dockerfile
@@ -61,6 +61,7 @@ RUN set -e; \
     && rm -rf "$GNUPGHOME" latest_toolchain.tar.gz.sig latest_toolchain.tar.gz
 
 ENV PATH="${SWIFT_PREFIX}/usr/bin:${PATH}"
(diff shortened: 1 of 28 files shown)


kkebo:
I updated this PR for 6.3 Dockerfiles.

kkebo:
If PR #590 is merged, this PR will be obsolete.

PR #479, opened 2025-06-20
Revert "Pin Windows images to the 6/9 nightly main snapshot"

owenv:
Reverts swiftlang/swift-docker#478

We've effectively worked around the issue that led to this pinning elsewhere, so I think it's best to unpin at this point.


owenv:
@shahmishal https://github.com/swiftlang/swift-corelibs-libdispatch/pull/883 fixed the underlying issue which forced us to pin this, so it should be safe to land this now

PR #483, opened 2025-06-30
Update patches and build the compiler from the latest commits, not an older tag

finagolfin:
Also, updated the C/C++ compilers used for Testing, as that was breaking the compiler-validated build.


finagolfin:
Sorry, wrong repo, this needs more testing.

finagolfin:
Please cancel the CI run of this pull, I don't have the permissions to do so.

PR #488, opened 2025-07-25
Update the Static SDK for Linux

al45tair:
Updated the versions of:

- libxml2 (2.14.5)
- curl (8.15.0)
- BoringSSL (newer SHA)

and added:

- bzip2 (1.0.8)
- xz-utils (5.8.1)
- libarchive (3.8.1)
- mimalloc (2.2.4)

Plus two security patches for musl to fix CVE-2025-26519.

Also link mimalloc by default, so programs using the Static SDK for Linux get a better memory allocator out of the box.

rdar://156423711

### DIFF
--- a/swift-ci/sdks/static-linux/resources/patches/musl/0001-iconv-fix-erroneous-input-validation-in-EUC-KR-decod.patch
+++ b/swift-ci/sdks/static-linux/resources/patches/musl/0001-iconv-fix-erroneous-input-validation-in-EUC-KR-decod.patch
@@ -1,0 +1,38 @@
+>From e5adcd97b5196e29991b524237381a0202a60659 Mon Sep 17 00:00:00 2001
+From: Rich Felker <dalias@aerifal.cx>
+Date: Sun, 9 Feb 2025 10:07:19 -0500
+Subject: [PATCH] iconv: fix erroneous input validation in EUC-KR decoder
+
+as a result of incorrect bounds checking on the lead byte being
+decoded, certain invalid inputs which should produce an encoding
+error, such as "\xc8\x41", instead produced out-of-bounds loads from
+the ksc table.
+
+in a worst case, the loaded value may not be a valid unicode scalar
+value, in which case, if the output encoding was UTF-8, wctomb would
+return (size_t)-1, causing an overflow in the output pointer and
+remaining buffer size which could clobber memory outside of the output
+buffer.
+
+bug report was submitted in private by Nick Wellnhofer on account of
+potential security implications.
+---
+ src/locale/iconv.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/locale/iconv.c b/src/locale/iconv.c
+index 9605c8e9..008c93f0 100644
+--- a/src/locale/iconv.c
++++ b/src/locale/iconv.c
+@@ -502,7 +502,7 @@ size_t iconv(iconv_t cd, char **restrict in, size_t *restrict inb, char **restri
+ 			if (c >= 93 || d >= 94) {
+ 				c += (0xa1-0x81);
+ 				d += 0xa1;
+-				if (c >= 93 || c>=0xc6-0x81 && d>0x52)
++				if (c > 0xc6-0x81 || c==0xc6-0x81 && d>0x52)
+ 					goto ilseq;
+ 				if (d-'A'<26) d = d-'A';
+ 				else if (d-'a'<26) d = d-'a'+26;
+-- 
+2.21.0
+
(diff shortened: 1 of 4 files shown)


al45tair:
This will require updates to the Static SDK build job, because it changes the versions of dependencies and adds some dependencies. Also, we will need to check the licensing situation before merging this.

shahmishal (approved):

al45tair:
Ideally we should land https://github.com/swiftlang/swift/pull/84705 and https://github.com/swiftlang/swift/pull/84706 first, as that will mean that all builds of the 0.1.0 version of the Static SDK for Linux will have library evolution disabled. That seems cleaner than landing this one first.

marcprux:
Tangential to this PR, but I'd like to align Android's dependencies with these. A few questions:

- Any particular reason for the specific `BORINGSSL_VERSION=817ab07ebb53da35afea409ab9328f578492832d`?
- Is the addition of the `bzip2`, `libarchive`, `mimalloc`, and `xz` dependencies somehow related to the version bumps for `boringssl`, `libcurl`, or `libxml2`?

al45tair:
> Any particular reason for the specific BORINGSSL_VERSION=817ab07ebb53da35afea409ab9328f578492832d?

Yes. I chose the hash Swift NIO is currently using. The problem with BoringSSL is that it doesn't actually have proper releases; the BoringSSL maintainers actively talk to the NIO maintainers and that's how _they_ get the hash to use. We should ideally pick up the same hash whenever they change it (it might be better to just check out the NIO repository to get the version we need, but that's not a change I've made here).

> Is the addition of the bzip2, libarchive, mimalloc, and xz dependencies somehow related to the version bumps for boringssl, libcurl, or libxml2?

No. Those were requests that I had from various parties: `mimalloc` because the Musl memory allocator is poor; `bzip2`, `libarchive` and `xz` because some folks wanted to use those compression libraries and I thought it made sense to make them part of the SDK itself, given how prevalent their use is.

marcprux:
Following-up on this, for the Android SDK we are just using the versions of curl and libxml2 that come along with the Swift checkouts (which I guess is coming from [update-checkout-config.json](https://github.com/swiftlang/swift/blob/652402e9e83d2e970de5a12c34e419f95d1cb7f3/utils/update_checkout/update-checkout-config.json#L189-L190))

Is there any particular reason to bump the versions of the curl and libxml2 dependencies for just the static SDK rather than for the Swift project as a whole? I ask because I'd also like a more recent version of libcurl for the Android SDK, but I am wondering if there will be concerns about not being feature/bug compatible with the dependencies across all the Swift platforms (e.g., Linux, BSD).

al45tair:
Interesting question. I think we should probably bump them for Swift more generally also, but since the Static SDK has its own copies, it seemed good to bump them here.

al45tair:
@shahmishal Can we get this merged? I don't have merge rights for `swift-docker`.

marcprux:
> I think we should probably bump them for Swift more generally also

I've filed a draft PR for this at https://github.com/swiftlang/swift/pull/86321

Cyberbeni:
@al45tair could these changes cause an increased memory usage? https://forums.swift.org/t/huge-memory-usage-increase-when-upgrading-from-6-2-3-to-6-2-4-with-static-linux-sdk/85719

al45tair:
> @al45tair could these changes cause an increased memory usage? https://forums.swift.org/t/huge-memory-usage-increase-when-upgrading-from-6-2-3-to-6-2-4-with-static-linux-sdk/85719

It's certainly _possible_. If you're seeing much higher memory usage, we could really do with knowing what's responsible for the allocations in question… not sure how easy that information would be to collect with the Static SDK though, as, well, it's all statically linked.

gregcotten:
hey @al45tair and @Cyberbeni, I've also been investigating increased memory usage of my Swift Static SDK executable since moving to 6.3.0. It seems that mimalloc v3 addresses some of these issues...

From https://github.com/microsoft/mimalloc:

> v3: recommended: simplifies the lock-free design of previous versions and improves sharing of memory between threads. On certain large workloads this version may use (much) less memory. Also supports true first-class heaps (that can allocate from any thread) and has more efficient heap-walking (for the CPython GC for example). (release tags: v3.x, development branch dev3).

Is there a reason the Static Linux SDK uses mimalloc v2? More stable possibly?

some v2 memory issues (there are more):
https://github.com/microsoft/mimalloc/issues/537
https://github.com/microsoft/mimalloc/issues/214

Cyberbeni:
@gregcotten Did you test mimalloc3 yet?

I have a script [here](https://codeberg.org/Cyberbeni/swift-builder-docker/src/branch/musl-allocator) on the `musl-allocator` branch for building the SDK without mimalloc, it's been working just as fine as with Swift 6.2.3 and earlier.

I've just updated it to build the SDK with mimalloc 3.4.4 on the `mimalloc3` branch.

Also have built versions [here](https://codeberg.org/Cyberbeni/-/packages/generic/swift_static_linux_sdk_musl_allocator/6.3.3), 0.0.2 is musl-allocator, 0.0.3 is mimalloc 3.4.4

And a builder docker image too [here](https://codeberg.org/Cyberbeni/-/packages/container/swift-builder/6.3.3-mimalloc3).

I didn't have time to try out the mimalloc3 version yet.

edit: tried the mimalloc3 version, without changing any settings through environment variables, it's still 69MB memory usage after starting the program that uses <5MB with Musl allocator. I think I'm going to stick to using the Musl allocator for my ~0 resources used services.

PR #493, opened 2025-08-12
Windows: Make sure the URL is fully formed before attempting to download

azharudd:
This should prevent the 404 errors when the URL is malformed due to `dir` and `download` fields missing.

### DIFF
--- a/nightly-main/windows/LTSC2022/Dockerfile
+++ b/nightly-main/windows/LTSC2022/Dockerfile
@@ -130,6 +130,14 @@ RUN Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:VSB});               
 # Install Swift toolchain.
 ARG SWIFT_RELEASE_METADATA=https://download.swift.org/development/windows10/latest-build.json
 RUN $Release = curl.exe -sL ${env:SWIFT_RELEASE_METADATA} | ConvertFrom-JSON;   \
+    Write-Host ('({0})' -f $Release);                                           \
+    Write-Host -NoNewLine 'Release metadata ... ';                              \
+    if ($Release.dir -and $Release.download) {                                  \
+      Write-Host '✓';                                                           \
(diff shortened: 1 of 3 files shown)


shahmishal (approved):

shahmishal:
Please also update other windows Dockerfiles too.

nightly-main/windows/LTSC2022/Dockerfile
```diff
@@ -130,6 +130,14 @@ RUN Write-Host -NoNewLine ('Downloading {0} ... ' -f ${env:VSB});
 # Install Swift toolchain.
 ARG SWIFT_RELEASE_METADATA=https://download.swift.org/development/windows10/latest-build.json
 RUN $Release = curl.exe -sL ${env:SWIFT_RELEASE_METADATA} | ConvertFrom-JSON;   \
+    Write-Host ('({0})' -f $Release);                                           \
```
compnerd, line 133: This will dump the json to the console - fine for debugging, but shouldn't be committed.

compnerd (changes requested):

nightly-main/windows/LTSC2022/Dockerfile
azharudd: @compnerd Are there any concerns about dumping the JSON file to the console? The JSON file is just a few lines long. I’m intentionally doing this so that if there are any failures, we can see what it contained and help us understand why it doesn’t contain the expected values.

azharudd:
> Please also update other windows Dockerfiles too.

Done!

nightly-main/windows/LTSC2022/Dockerfile
compnerd: It makes it harder to see what is going on. If you want to do this at the end of the run if there are failures, that is fine, but we shouldn't do this in all cases IMO.

PR #501, opened 2025-09-22
[Static SDK] Build clang before running the main build.

al45tair:
We need to do this because otherwise libc++ fails to build, since the clang we pick up from the Swift install is an older version than the one libc++ requires (libc++ currently needs Clang 19; we ship Clang 17 with Swift).

### DIFF
--- a/swift-ci/sdks/static-linux/scripts/build.sh
+++ b/swift-ci/sdks/static-linux/scripts/build.sh
@@ -268,6 +268,37 @@ else
     exit 1
 fi
 
+# -----------------------------------------------------------------------
+
+header "Building clang for host"
+
+mkdir -p ${build_dir}/host/clang ${build_dir}/clang
+
+run cmake -G Ninja -S ${source_dir}/swift-project/llvm-project/llvm \
+    -B ${build_dir}/host/clang \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_ENABLE_PROJECTS="clang" \
+    -DLLVM_PARALLEL_LINK_JOBS=1 \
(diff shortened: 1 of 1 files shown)


al45tair:
(I'm still testing this locally, but I think this should fix the problem that we're seeing.)

al45tair:
It turns out that BoringSSL's `-Werror` causes a build failure if you use a newer compiler. So we need to turn that off. Still testing.

al45tair:
This built successfully locally, with the rebranch checkout.

swift-ci/sdks/static-linux/scripts/build.sh
```diff
@@ -268,6 +272,45 @@ else
...
+echo "done"
+
+# -----------------------------------------------------------------------
+
+header "Building clang for host"
+
+mkdir -p ${build_dir}/host/clang ${build_dir}/clang
+
+run cmake -G Ninja -S ${source_dir}/swift-project/llvm-project/llvm \
+    -B ${build_dir}/host/clang \
```
AnthonyLatsis, line 290: Why does this work with the source dir being `llvm-project/llvm` instead of `llvm-project/clang`?

shahmishal (approved):

swift-ci/sdks/static-linux/scripts/build.sh
al45tair: It actually _doesn't_ work with the source dir being `llvm-project/clang` (that's the first thing I tried). Using `llvm` rather than `clang` is actually [what the instructions here say to do](https://clang.llvm.org/get_started.html).

PR #505, opened 2025-10-16
Windows: add Dockerfile to smoke test arbitrary installers

speednoisemovement:
Split off from https://github.com/swiftlang/swift/pull/84770

This adds a Dockerfile which installs a given Swift installer and sets the given script to the entry point.


shahmishal:
@speednoisemovement Can you also create a match Dockerfile for 2019?

speednoisemovement:
> @speednoisemovement Can you also create a match Dockerfile for 2019?

Done. IIUC 1809 is the same ltsc2019 and it's what's used elsewhere in this repo, please LMK if that's wrong.

justice-adams-apple:
@speednoisemovement 

Could we enable long paths so we don't run into any issues with the OS file system
```
reg add "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem" 
  /v "LongPathsEnabled" /d "1"
  ```
  
  And also could we install python for lldb

compnerd:
> @speednoisemovement
> 
> Could we enable long paths so we don't run into any issues with the OS file system
> 
> ```
> reg add "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem" 
>   /v "LongPathsEnabled" /d "1"
> ```

I would really prefer that we do not do this. This is going to hide issues with long paths that users might run into. We cannot assume that the user can set this (corporate environments do often block the ability to set arbitrary HKLM nodes).

> And also could we install python for lldb

We should be able to do this at least.

justice-adams-apple:
For 2022 we should use Python 3.9.13

For 1809 we should use Python 3.7.8

That way the lldb testing uses the same python that built the installer

speednoisemovement:
> For 2022 we should use Python 3.9.13
> 
> For 1809 we should use Python 3.7.8
> 
> That way the lldb testing uses the same python that built the installer

Will do, but can you point me to where those versions are coming from?

justice-adams-apple:
They are from our CI machines which will build the installer in our CI jobs, someone at desk could use an arbitrary python but since these dockerfiles are for CI we should likely match the builders

justice-adams-apple (approved):
LGTM, thank you!

Issue #60, opened 2017-01-27
3.1 Bug Workaround


swizzlr:
Add this to your Dockerfile to remove the offending code that causes the 3.1 image to fail.
RUN sed -i '/#include <Block.h>/d' /usr/lib/swift/CoreFoundation/CFBase.h

Tracking the issue in https://github.com/apple/swift-corelibs-foundation/pull/843

PR #507, opened 2025-10-29
Android: avoid compiler issue swiftlang/swift#55637 when building the stdlib, where it's picking up a module from another arch

finagolfin:
Hopefully, this stops the random flakes we've been seeing on the Android CI, tested by swift-android-sdk/swift-docker#10 externally first.

### DIFF
--- a/swift-ci/sdks/android/scripts/build.sh
+++ b/swift-ci/sdks/android/scripts/build.sh
@@ -498,10 +498,11 @@ for arch in $archs; do
             --foundation-cmake-options=-DCMAKE_SHARED_LINKER_FLAGS= \
             --cross-compile-append-host-target-to-destdir=False 
             # --extra-cmake-options='-DCMAKE_EXTRA_LINK_FLAGS="-Wl,-z,max-page-size=16384"'
-        # need to remove symlink that gets created in the NDK to the previous arch's build
-        # or else we get errors like:
+        # need to remove the arch-specific portion of the Swift resource dir in
+        # the build directory, through this symlink that we create in the NDK to
+        # the build directory, or else we get errors like:
         # error: could not find module '_Builtin_float' for target 'x86_64-unknown-linux-android'; found: aarch64-unknown-linux-android, at: /home/runner/work/_temp/swift-android-sdk/ndk/android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/swift/android/_Builtin_float.swiftmodule
-        rm -f $ndk_installation/sysroot/usr/lib/swift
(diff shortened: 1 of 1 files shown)


finagolfin:
@shahmishal, this should avoid [the flakes we've been seeing on the Android CI](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/49/console), by deleting this build directory for each Android arch after it has been installed.

shahmishal (approved):

swift-ci/sdks/android/scripts/build.sh
```diff
@@ -498,10 +498,11 @@ for arch in $archs; do
...
             --cross-compile-append-host-target-to-destdir=False 
             # --extra-cmake-options='-DCMAKE_EXTRA_LINK_FLAGS="-Wl,-z,max-page-size=16384"'
-        # need to remove symlink that gets created in the NDK to the previous arch's build
-        # or else we get errors like:
+        # need to remove the arch-specific portion of the Swift resource dir in
+        # the build directory, through this symlink that we create in the NDK to
+        # the build directory, or else we get errors like:
         # error: could not find module '_Builtin_float' for target 'x86_64-unknown-linux-android'; found: aarch64-unknown-linux-android, at: /home/runner/work/_temp/swift-android-sdk/ndk/android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/swift/android/_Builtin_float.swiftmodule
-        rm -f $ndk_installation/sysroot/usr/lib/swift
+        rm -rf $ndk_installation/sysroot/usr/lib/swift/android
```
marcprux, line 505: Is this right? Elsewhere we run `ln -s $ndk_installation/lib/clang/$ndk_clang_version $swift_res_root/usr/lib/swift/clang`, so maybe we should be deleting from `$swift_res_root` instead?

Maybe print out the path that is being delete here, and cross reference it with the error message failures like in [CI run #60](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/60/console):

```
02:14:31  <unknown>:0: error: could not find module '_math' for target 'x86_64-unknown-linux-android'; found: aarch64-unknown-linux-android, at: /home/build-user/build/swift-project/Ninja-Release/swift-linux-x86_64/lib/swift/android/_math.swiftmodule/x86_64-unknown-linux-android
```

finagolfin: Yeah, that is what I was worried about too, that deleting these files in the build directory through the symlink may not work, but [as I mentioned earlier](https://github.com/swiftlang/swift-docker/issues/510#issuecomment-3490508364), it does seem to work.

I will look into the underlying compiler issue this week and try to fix or work around the root cause of these flakes.

PR #455, opened 2025-03-30
Install binaries under `$SWIFT_PREFIX/usr/bin` instead of `/usr/bin`

kkebo:
## Motivation

fixes #383 

This prevents some Swift toolchain binaries from being unintentionally overwritten by system package managers (e.g. apt).

## Changes

I added the following changes to 6.0 Dockerfiles, 6.1 Dockerfiles, nightly-6.1 Dockerfiles, nightly-6.2 Dockerfiles, and nightly-main Dockerfiles.

- Extract a tarball into `$SWIFT_PREFIX` instead of `/`
  - `$SWIFT_PREFIX` is `/opt/swift/{version number}` by default.
- Set `PATH="${SWIFT_PREFIX}/usr/bin:${PATH}"` and `LD_LIBRARY_PATH="${SWIFT_PREFIX}/usr/lib/swift/linux"`
- Create symbolic links of `$SWIFT_PREFIX/usr/bin/*` to `/usr/bin/*` for existing users who expect `swift` to be in `/usr/bin`.

PR #516, opened 2025-12-08
Update `SWIFT_VERSION` of nightly-main images to 6.3

kkebo:
The nightly-main images should be now 6.3, so 6.2 is wrong. This change is needed only for Debian and Fedora images since the other images don't have `SWIFT_VERSION` until #455 is merged.


bstan7168-rgb:
Thx

bstan7168-rgb:
Thx

bstan7168-rgb (approved):

PR #527, opened 2026-02-10
[source compatibility] add windows CI dockerfile

justice-adams-apple:
Add a Dockerfile with all the dependencies needed to install an arbitrary already-built swift compiler and run the source compat suite on windows.

This is only for use by CI jobs. I will be adding an equivalent linux Dockerfile to the `swift-ci/source-compatibility/` folder next so our nightly compat jobs build the toolchain from ToT

### DIFF
--- a/swift-ci/source-compatibility/windows/1809/Dockerfile
+++ b/swift-ci/source-compatibility/windows/1809/Dockerfile
@@ -1,0 +1,151 @@
+FROM mcr.microsoft.com/windows/servercore:1809 AS windows
+
+LABEL maintainer="Swift Infrastructure <swift-infrastructure@forums.swift.org>"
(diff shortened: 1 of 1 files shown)


compnerd:
How is this any different from the existing docker images for Windows? The only ones that are different are the ones for building the toolchain, but this seems identical to the other docker images.

justice-adams-apple:
@compnerd this specifically installs from a local installer and its 1809 based (required for our current infra). So it allows us to build the compiler locally and then build the container image containing said compiler rather than pulling and installing a toolchain package/installer from download.swift.org

justice-adams-apple:
I'm going to close this in favor of
https://github.com/swiftlang/swift-docker/pull/505

PR #537, opened 2026-03-25
Do not create dangling symlink in Android SDK artifactbundle

### LINKED ISSUE #535: Swift 6.3 SDK for Android has dangling clang symlinks with absolute paths
I installed the Android SDK from the official 6.3 release and found this:

 Directory of C:\Users\jakepetroules\.swiftpm\swift-sdks\swift-6.3-RELEASE_android.artifactbundle\swift-android\swift-resources\usr\lib\swift

03/20/2026  09:17 PM    <DIR>          .
03/20/2026  09:17 PM    <DIR>          ..
03/21/2026  12:37 AM    <SYMLINK>      clang [\home\build-user\build\ndk\android-ndk-r27d\toolchains\llvm\prebuilt\linux-x86_64\lib\clang\18]

That symlink must be a side effect of the build process, but it needs to be deleted because we shouldn't be shipping broken symlinks.

marcprux:
Fixes https://github.com/swiftlang/swift-docker/issues/535

### DIFF
--- a/swift-ci/sdks/android/scripts/build.sh
+++ b/swift-ci/sdks/android/scripts/build.sh
@@ -678,9 +678,9 @@ done
 
 rm -rf ${swift_res_root}/usr/share/{aclocal,doc,man}
 rm -r ${sdk_staging}
-# Link the local NDK's clang resource directory as a place-holder
+# The local NDK's clang resource directory will be linked under
+# here as part of the post-install script
 mkdir -p $swift_res_root/usr/lib/swift
-ln -s $ndk_installation/lib/clang/$ndk_clang_version $swift_res_root/usr/lib/swift/clang
 
 # create an install script to set up the NDK links
 #ANDROID_NDK_HOME="/opt/homebrew/share/android-ndk"


finagolfin:
Have you tested building with this locally?

marcprux:
> Have you tested building with this locally?

No. TBH, I was hoping to get this in and then test the resulting .artifactbundle that gets produced…

finagolfin:
OK, let me test it locally, then we can get this in. No hurry, as the next patch release likely won't be till next month.

jakepetroules:
Thanks for the quick fix. However, please note that there are _multiple_ dangling symlinks:

```
$ find . -type l
./swift-android/swift-resources/usr/lib/swift_static-aarch64/clang
./swift-android/swift-resources/usr/lib/swift-armv7/clang
./swift-android/swift-resources/usr/lib/swift_static-armv7/clang
./swift-android/swift-resources/usr/lib/swift-x86_64/clang
./swift-android/swift-resources/usr/lib/swift/clang
./swift-android/swift-resources/usr/lib/swift_static-x86_64/clang
./swift-android/swift-resources/usr/lib/swift-aarch64/clang
```

`./swift-android/swift-resources/usr/lib/swift/clang` is the one that points to the absolute path, while the other 6 are pointing to `../swift/clang`, so they themselves will become dangling once this one is removed.

Can you remove the others as well and have the script create them on demand?

Since the script is no longer needed with SwiftPM's Swift Build backend, I'd like to avoid baking in assumptions that the script _will_ be run (and then we can simply remove the script in 6.5 or 6.6 or whenever we end up removing the legacy build backend).

marcprux:
> Can you remove the others as well and have the script create them on demand?

Ahh, I thought only the absolute symlink was the issue.

I agree that we should move that link creation over to the post-install script, but that will require a bit more careful work, since those dangling symlinks are currently being "filled in" by the post-install script. We will instead need to make the script perform the link creation instead.

> (and then we can simply remove the script in 6.5 or 6.6 or whenever we end up removing the legacy build backend)

I look forward to that future 😶

finagolfin:
Oh, I think I remember now why we put it in: we added the other symlinks to this central one, then archiving the bundle failed or something if they were all pointing at a non-existent symlink, so I put this in as a place-holder.

I will confirm that is the reason and let you know, but either way, we'll likely need to ship with all of them, or add all of them after installation.

finagolfin:
I'm building locally now without that last symlink, will let you know what I find.

finagolfin:
Forgot about this as bigger stuff came up... the current plan is to ditch this install script altogether with the 6.4 release, as `swift sdk configure` now works in trunk and swift-build has other ways of working around these issues. I'm not sure these dangling symlinks really matter in 6.3 alone in the meantime.

I could not reproduce the build error I saw before on CI, when building locally now outside Docker, but that doesn't mean the CI won't still hit it. If you still want to clean this up, maybe replace all the symlinks with symlinks created in the install script when run by the user? No symlinks in the bundle means it won't fail when building, and Jake's dangling symlink concern would be addressed..

justice-adams-apple (approved):

PR #544, opened 2026-04-21
Install unzip in the Ubuntu nightlies

FranzBusch:


justice-adams-apple:
could you add this to the buildx files as well
* https://github.com/swiftlang/swift-docker/blob/main/nightly-main/ubuntu/24.04/buildx/Dockerfile
* https://github.com/swiftlang/swift-docker/blob/main/nightly-6.3/ubuntu/24.04/buildx/Dockerfile

Would it make sense to include this change to the 22.04 Dockerfiles as well?

FranzBusch:
> Would it make sense to include this change to the 22.04 Dockerfiles as well?

The 22.04 seem to have this already by installing `zip` which also installs `unzip`.

justice-adams-apple:
@FranzBusch ahh ok makes sense, I guess maybe a follow up for this change then is: should we use zip?

justice-adams-apple (approved):
Pending question about zip vs unzip but otherwise LGTM

FranzBusch:
> @FranzBusch ahh ok makes sense, I guess maybe a follow up for this change then is: should we use zip?

I think it would be good to align this but I didn't wanna touch any Dockerfiles here that are not showing a problem. Feel free to follow up and clean this up across all the files.

PR #550, opened 2026-04-29
[Ubuntu 26.04] Fix libxml linking path

justice-adams-apple:


shahmishal (approved):
We should revert this once the first toolchain is available.

PR #556, opened 2026-05-21
[android] Add option to call fetch-source with --github-comment

justice-adams-apple:
Allow for calling fetch-source with `--github-comment`. We use `fetch-source` when running PR tests so this will allow us to eventually pass GH comments to update-checkout from jobs which trigger android PR CI


justice-adams-apple:
cc @finagolfin

finagolfin:
Thanks, looking forward to using it.

PR #557, opened 2026-05-22
[Static SDK] Disable assertions in the Swift build.

al45tair:
We were inadvertently building with assertions enabled.  Turn them off for better performance.

rdar://177470294

### DIFF
--- a/swift-ci/sdks/static-linux/scripts/build.sh
+++ b/swift-ci/sdks/static-linux/scripts/build.sh
@@ -115,7 +115,7 @@ declare_package libarchive "libarchive" "BSD-2-Clause" "https://www.libarchive.o
 declare_package mimalloc "mimalloc" "MIT" "https://microsoft.github.io/mimalloc/"
 
 # Parse command line arguments
-static_linux_sdk_version=0.1.0
+static_linux_sdk_version=0.1.1
 sdk_name=
 archs=x86_64,aarch64
 build_type=RelWithDebInfo
@@ -855,7 +855,7 @@ EOF
 
     SWIFT_SOURCE_ROOT="${source_dir}/swift-project" \
     SWIFT_BUILD_ROOT="${build_dir}/swift" \
-    run ${source_dir}/swift-project/swift/utils/build-script -r \
+    run ${source_dir}/swift-project/swift/utils/build-script -r -A \
         --reconfigure \
         --compiler-vendor=apple \
         --bootstrapping hosttools \
--- a/swift-ci/sdks/static-linux/scripts/build.sh
+++ b/swift-ci/sdks/static-linux/scripts/build.sh
@@ -115,7 +115,7 @@ declare_package libarchive "libarchive" "BSD-2-Clause" "https://www.libarchive.o
 declare_package mimalloc "mimalloc" "MIT" "https://microsoft.github.io/mimalloc/"
 
 # Parse command line arguments
-static_linux_sdk_version=0.1.1
+static_linux_sdk_version=0.1.0
 sdk_name=
 archs=x86_64,aarch64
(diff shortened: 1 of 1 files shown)


al45tair:
**Note: this also bumps the Static SDK version to 0.1.1; any scripts that hard-code the version number will need to change.**

al45tair:
(Build failure is nothing to do with this PR.)

swift-ci/sdks/static-linux/scripts/build.sh
```diff
@@ -855,7 +855,7 @@ EOF
 
     SWIFT_SOURCE_ROOT="${source_dir}/swift-project" \
     SWIFT_BUILD_ROOT="${build_dir}/swift" \
-    run ${source_dir}/swift-project/swift/utils/build-script -r \
+    run ${source_dir}/swift-project/swift/utils/build-script -r -A \
```
shahmishal, line 858: We should create a preset

shahmishal (changes requested):

swift-ci/sdks/static-linux/scripts/build.sh
```diff
@@ -115,7 +115,7 @@ declare_package libarchive "libarchive" "BSD-2-Clause" "https://www.libarchive.o
 declare_package mimalloc "mimalloc" "MIT" "https://microsoft.github.io/mimalloc/"
 
 # Parse command line arguments
-static_linux_sdk_version=0.1.0
```
justice-adams-apple, line 118: Feel free to ping me when this lands and I'll make sure the mainline CI job is updated to build the correct version

shahmishal: Can we drop the version? It creates more work for no added value.

FranzBusch: +1

al45tair: > It creates more work for no added value.

How are people to know whether (for instance) they've got a build of the Static SDK with this fix if we don't version it?

Equally, if we update the dependencies, how do they know they've got a version with the updated dependencies? (I guess for this they could check the SBOM, since we do have one).

We could drop the version from the filename maybe? That does make life harder for people though, if they care about some specific change we make, since Swift SDKs don't have a version field or any way to ask SwiftPM to tell you what version you have.

(Maybe that should be the ask? Add a version field to the Swift SDK support in SwiftPM, and list the version when you do `swift sdk list`?)

swift-ci/sdks/static-linux/scripts/build.sh
al45tair: We _could_ do that, but then the build settings for the SDK are separate from the SDK's own build script, and then we have a versioning problem, since we could update the build settings and end up with a build of the Static SDK with the same version number as another build, but that's built with different build settings.

This wouldn't be a problem if the build scripts lived in the Swift repo rather than the swift-docker repo. Do we want to do that instead?

swift-ci/sdks/static-linux/scripts/build.sh
al45tair: Or maybe we could update the CI script to look for a file with *any* version number as the output, and if it needs the version number for some reason then extract it from the filename?

swift-ci/sdks/static-linux/scripts/build.sh
```diff
@@ -115,7 +115,7 @@ declare_package libarchive "libarchive" "BSD-2-Clause" "https://www.libarchive.o
...
+static_linux_sdk_version=0.1.1
```
shahmishal:
```suggestion
static_linux_sdk_version=0.1.0
```

PR #562, opened 2026-05-30
android: Use Android NDK r28 default for 16kb supporting

utzcoz:
See
https://developer.android.com/guide/practices/page-sizes#compile-16-kb-alignment.

### DIFF
--- a/swift-ci/sdks/android/build-docker
+++ b/swift-ci/sdks/android/build-docker
@@ -17,7 +17,7 @@
 # default architectures to build for
 TARGET_ARCHS=${TARGET_ARCHS:-aarch64,x86_64,armv7}
 
-ANDROID_NDK_VERSION=android-ndk-r27d
+ANDROID_NDK_VERSION=android-ndk-r28c
 ANDROID_API=28
 
 BASEPATH=$(dirname $(realpath $0))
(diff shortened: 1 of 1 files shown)


finagolfin:
What problem are you facing exactly? We already switched all Android libraries in the SDK to use 16 KB page alignment last summer in swiftlang/swift#81596, and 6.4 and trunk have been switched to NDK 28c, soon the next LTS NDK 30 beta.

We only keep the current Swift 6.3 release building with NDK 27, which should work fine to set this page alignment provided you pass in the correct flag, as we're doing already.

utzcoz:
I don't notice flags in swift. If they are added into swift building flags, it's okay to keep using NDK r27. The main reason of upgrading to NDK r28 is to enable 16kb supporting without flags.

finagolfin:
The flags should be added in automatically by our SDK bundle config. If you run this command, it will show you the path to the toolset file that applies extra build flags to every Android build you run:
```
> swift sdk configure --show-configuration swift-6.3.2-RELEASE_android
--- snip lots of repetitive output ---
toolsetPaths: ["/home/finagolfin/.swiftpm/swift-sdks/swift-6.3.2-RELEASE_android.artifactbundle/swift-android/swift-toolset.json"]

> cat /home/finagolfin/.swiftpm/swift-sdks/swift-6.3.2-RELEASE_android.artifactbundle/swift-android/swift-toolset.json                                                              {
  "cCompiler": { "extraCLIOptions": ["-fPIC"] },
  "swiftCompiler": { "extraCLIOptions": ["-Xclang-linker", "-fuse-ld=lld"] },
  "linker": { "extraCLIOptions": ["-z", "max-page-size=16384"] },
  "schemaVersion": "1.0"
}
```
You can check that the flag is applied properly by running `swift build` with the verbose `-v` flag, to dump all the flags used, and of course by running the resulting libraries through the `check_elf_alignment.sh` script from your link above.

If you see any issues, let us know, but this pull won't help with that.

PR #566, opened 2026-06-03
[windows][lldb] reduce Docker image size

charles-zablit:
This patch reduces the resulting image size:

1. Switch from `vs_community.exe` to `vs_buildtools.exe`: Build Tools is headless, no IDE.
2. `--depth 1` on vcpkg clone to avoid fetching the full git history.
3. Delete `vcpkg` buildtrees, packages, downloads after the libxml2 install.

Overall, this should reduce the image size from 9GB to ~1GB.

### DIFF
--- a/swift-ci/main/windows/lldb/1809/Dockerfile
+++ b/swift-ci/main/windows/lldb/1809/Dockerfile
@@ -9,9 +9,9 @@ RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tl
 RUN Set-ItemProperty -Path HKLM:\\SYSTEM\\CurrentControlSet\\Control\\FileSystem -Name LongPathsEnabled -Value 1
 RUN Set-ItemProperty -Path HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\AppModelUnlock -Name AllowDevelopmentWithoutDevLicense -Value 1 -Type DWord
 
-# Visual Studio 2022 Community with C++ and Windows SDK components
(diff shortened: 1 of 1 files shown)


swift-ci/main/windows/lldb/1809/Dockerfile
```diff
@@ -60,14 +60,15 @@ RUN git config --global core.symlink true; \
     git config --global core.autocrlf false
 
 # vcpkg + libxml2
-RUN git clone https://github.com/microsoft/vcpkg C:\\vcpkg; \
+RUN git clone --depth 1 https://github.com/microsoft/vcpkg C:\\vcpkg; \
     C:\\vcpkg\\bootstrap-vcpkg.bat -disableMetrics
 SHELL ["cmd", "/S", "/C"]
-RUN ""C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat" && C:\vcpkg\vcpkg install libxml2:x64-windows"
+RUN ""C:\Program Files\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvars64.bat" && C:\vcpkg\vcpkg install libxml2:x64-windows"
```
compnerd, line 66: Why do we need this at all? We build libxml2 from source for the toolchain, we should just do that here and avoid vcpkg entire.

charles-zablit: We could yes. I'm not able to test this at desk, so I wanted to minimize the changes to avoid any breakage.

PR #569, opened 2026-06-09
Don't build the embedded stdlib or copy the trunk SwiftBridging header on Android CI

finagolfin:
The Android CI is [broken right now only because we're building the embedded stdlib unnecessarily](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/503/console), and that is currently broken on most linux toolchain CI, swiftlang/swift#89753. Disable it as [we just throw out the results](https://github.com/swiftlang/swift-docker/blob/05c46ef7c5867f040460def72f1497b76af32d71/swift-ci/sdks/android/scripts/build.sh#L664) anyway.

Let me test this locally and it should be ready to go.

### DIFF
--- a/swift-ci/sdks/android/scripts/build.sh
+++ b/swift-ci/sdks/android/scripts/build.sh
@@ -496,6 +496,7 @@ for arch in $archs; do
             --xctest --install-xctest \
             --swift-testing --install-swift-testing \
             --swift-testing-macros --install-swift-testing-macros \
+            --build-embedded-stdlib=False \
             --cross-compile-build-swift-tools=False \
             --libdispatch-cmake-options=-DCMAKE_SHARED_LINKER_FLAGS= \
             --foundation-cmake-options=-DCMAKE_SHARED_LINKER_FLAGS= \
--- a/swift-ci/sdks/android/scripts/build.sh
+++ b/swift-ci/sdks/android/scripts/build.sh
@@ -654,7 +654,10 @@ for arch in $archs; do
     quiet_pushd ${sdk_staging}/${arch}/usr
         rm -rf bin lib/clang local
         rm -r include/*
-        cp -r ${swift_source_dir}/swift/lib/ClangImporter/SwiftBridging/{module.modulemap,swift} include/
+        # No longer needed in trunk snapshots and PRs
+        if [[ $swift_version != DEVELOPMENT-SNAPSHOT-* && $swift_version != PR ]]; then
+            cp -r ${swift_source_dir}/swift/lib/ClangImporter/SwiftBridging/{module.modulemap,swift} include/
+        fi
 
         arch_triple="$arch-linux-android"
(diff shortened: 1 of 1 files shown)


DougGregor:
I think https://github.com/swiftlang/swift/pull/89753 will unbreak the CI. My other PR doesn't have this fix.

finagolfin:
I didn't even really notice we're building embedded till now, so it was pretty stable, :+1: just a waste to keep building it and then throwing it out. If we ever ship an embedded subset for Android, we'll enable this again.

finagolfin:
Alright, worked fine locally, ready for review.

finagolfin:
Added another commit to fix [a CI failure when bundling](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/511/console), ready to go in, @justice-adams-apple.

finagolfin:
Updated the bridging fix so it should only affect trunk, where this header was moved, swiftlang/swift#88748.

finagolfin:
@shahmishal, this will fix the Android CI, ready to go.

finagolfin:
Noticed that this would likely still break for [the less commonly-run trunk PR CI](https://ci.swift.org/job/swift-PR-swift-sdk-for-android/79/console), so updated with a fix for that too.

DougGregor:
I want to poke at the "why?" here a little more. Embedded can be useful even for non-embedded targets because it doesn't require a runtime underneath, so you get independent binaries with smaller code size. I think Android users can benefit from that as well.

finagolfin:
Sure, the problem is that nobody that I know of has gotten the Embedded subset compiled for Android, so [the flag that would enable it is disabled by default](https://github.com/swiftlang/swift/blob/f38c3c7d7ef0366d5c8f6cb67b1f887c7320c830/utils/build_swift/build_swift/driver_arguments.py#L1295) and isn't wired up to build for Android anyway. However, since this `--build-embedded-stdlib` flag is enabled by default, we unnecessarily build the Embedded subset for linux/glibc alone on the Android CI, only to throw it out when creating the final Android bundle. All this pull does is stop wasting time building that linux/glibc Embedded stdlib on the Android CI.

As for getting the Embedded subset for Android going, there was broad agreement that it would be useful when the @swiftlang/android-workgroup discussed it a couple times last year, but nobody has stepped up to build it, or seemingly even tried to, as nobody brought it up this year.

I think we all agree that we'd like to have it, but nobody has stepped up to provide it, while this pull has no effect on Embedded Android.

finagolfin:
In case it is unclear what I'm talking about, take a look at the failing build command in my first link in the OP, where the Embedded stdlib for linux-gnu was failing. We never built the Embedded stdlib for Android on the Android CI, so this pull simply disables building that linux-gnu Embedded stdlib alone.

shahmishal:
Going to merge this to unblock Android CI.

PR #429, opened 2024-11-11
Configurable user IDs

etcwilde:
When working with these containers to debug issues locally, it's sometimes useful to have the source directory live on the host, while building in the container. In order for permissions to line up, the user ID numbers need to match or the build-user may not be able to create the necessary directories.

This patch allows setting the build-user user ID so that the permissions line up and the user in the container and host can manipulate the files.

Override the build user ID with the current user's ID is done with the following command to ensure that the permissions on any volumes are kept in sync is the following:
`docker build --build-arg BUILD_USER_ID=$(id -u) ...`

PR #572, opened 2026-06-22
UBI9: Configurable build user id

etcwilde:
Making the builder-user ID configurable through the docker invocation means that folks can use volumes to map an existing Swift checkout on their host into the container to build in this environment without running into issues with permissions in the build directory.

e.g. `docker build --build-arg BUILD_USER_ID=$(id -u) ...`

By default, the `BUILD_USER_ID` is `42`, matching the existing setup.


etcwilde:
Closing in favor of https://github.com/swiftlang/swift-docker/pull/429.

PR #573, opened 2026-06-22
[windows][lldb] fix Visual Studio path

charles-zablit:
The Visual Studio path is incorrect, causing the build to fail.

This patch fixes it.


shahmishal (approved):

compnerd:
The path was correct - we should be using 64-bit visual studio, not 32-bit.

PR #578, opened 2026-07-02
Fix various issues with the Static Linux SDK build script

Cyberbeni:
All the changes should be self-explanatory
cc @al45tair


Cyberbeni:
ping @shahmishal @justice-adams-apple @al45tair

PR #581, opened 2026-07-20
[ubuntu 2604] use bsdtar on 26.04

justice-adams-apple:
working around an issue with tar in 26.04.

Shown to work in CI


shahmishal (approved):

justice-adams-apple:
Only failing 6.4x until the toolchain is set up, should be good to go

PR #582, opened 2026-07-22
[android] Build Android Embedded Swift stdlib

finnvoor:
Adds an `-embedded` variant of the android sdk.

Depends on https://github.com/swiftlang/swift/pull/90768 and https://github.com/swiftlang/swift/pull/90886

cc @finagolfin

### DIFF
--- a/swift-ci/sdks/android/scripts/build.sh
+++ b/swift-ci/sdks/android/scripts/build.sh
@@ -496,7 +496,6 @@ for arch in $archs; do
             --xctest --install-xctest \
             --swift-testing --install-swift-testing \
             --swift-testing-macros --install-swift-testing-macros \
-            --build-embedded-stdlib=False \
             --cross-compile-build-swift-tools=False \
             --libdispatch-cmake-options=-DCMAKE_SHARED_LINKER_FLAGS= \
             --foundation-cmake-options=-DCMAKE_SHARED_LINKER_FLAGS= \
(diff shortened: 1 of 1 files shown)


finagolfin:
Thanks, have you tried building this yourself and using the resulting SDK bundle?

finagolfin:
Never mind, see that you did.

finagolfin:
I wonder if we should register the embedded SDK as a separate SDK bundle, like I think wasm does. That way, there is no collision between the target triples.

finnvoor:
Makes sense, tried that and tested locally and the SDK layout works (https://github.com/finnvoor/swift-docker/commit/9df7edc0dd8618d500ef82c14f502eb39a9ba684), but I ran into an Android-specific Swift Build issue: the Android platform plugin currently overrides SDK toolsets with [`overrideToolsetAbsolutePaths: []`](https://github.com/swiftlang/swift-build/blob/7f96ee0e8aa8a35fe84f3e80932eb9e456356d10/Sources/SWBAndroidPlatform/Plugin.swift#L173-L180), so the `embedded-toolset.json` from the `-embedded` SDK is ignored. Passing the same toolset explicitly works, so the packaging is fine, but the SDK-bundled toolset is not applied for Android yet. So just running `swift build --swift-sdk <sdk-id>_android-embedded --triple aarch64-unknown-linux-android23` doesn't apply the embedded toolset yet.

finagolfin:
Yeah, that is a known issue that I just discussed with the author of that internal build logic: I will try to re-enable toolsets for Android soon, then we'll see if this works with your new config. Meanwhile, keep this pull ready with whatever config you need, don't see any new toolset added here yet?

finnvoor:
sounds good, PR updated with standalone embedded sdk. Everything should work once the toolset makes it through swift build.

PR #585, opened 2026-08-07
Debian 10: Update apt sources.list file

etcwilde:
Debian 10 packages are no longer at the 'http://deb.debian.org' path as the distro is very old. Packages are now located at 'http://archive.debian.org'. Updating the `sources.list` file in the image to point at the appropriate location.


shahmishal:
We should drop Debian 10, its no longer supported

etcwilde:
Hmm, CI does not like it when I remove that file.

PR #586, opened 2026-08-07
Static Linux: Add LLVM_TARGETS_TO_BUILD to set architectures explicitly

MaxDesiatov:
We're not supposed to build with AMDGPU LLVM backend here, otherwise this leads to unsupported code paths:

```
 /home/build-user/source/swift-project/llvm-project/llvm/lib/Target/AMDGPU/AMDGPUInstCombineIntrinsic.cpp:1211: error: invalid conversion from 'llvm::Value*' to 'llvm::CallInst*' [-fpermissive]
```

rdar://184231422


shahmishal:
cc: @al45tair

